PR #7, opened 2015-01-13
Add new prop type for bounds

andywer:
Adds a new PropTypes type. ImageOverlay, Map and Reactangle allow a `bounds` property to be set whose value might either be an array of LatLng values or (and that's fixed by this pull request) an instance of Leaflet.LatLngBounds.

Previous implementation caused warnings on the console if a (valid) LatLngBounds instance was passed.

See http://leafletjs.com/reference.html#map-maxbounds

### DIFF
--- a/example/build/lib.js
+++ b/example/build/lib.js
@@ -47,7 +47,7 @@ module.exports = React.createClass({
   }
 });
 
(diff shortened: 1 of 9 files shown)


PaulLeCam:
Thanks!

PR #8, opened 2015-01-14
Fix popup loosing options on `popupContainer.bindPopup()`

andywer:
This commit fixes an issue when popups could suddenly lose their options. That's because `popupContainer.bindPopup()` actually creates a new popup. Now the popup options are passed to `bindPopup`.


PaulLeCam:
Cool, thanks!

PR #9, opened 2015-01-15
Allow to change the leaflet imageOverlay.

scherler:
Hi, I have the need to change the url and some other props on the ImageOverlay, so I implemented the componentWillReceiveProps and then either create a new overLay or reset the old one (after changing the props).

### DIFF
--- a/lib/ImageOverlay.js
+++ b/lib/ImageOverlay.js
@@ -13,26 +13,18 @@ module.exports = React.createClass({
     url: React.PropTypes.string.isRequired,
     bounds: boundsType.isRequired
   },
-  setInitial: function(properties){
-    var $__0=     properties,bounds=$__0.bounds,map=$__0.map,url=$__0.url,props=(function(source, exclusion) {var rest = {};var hasOwn = Object.prototype.hasOwnProperty;if (source == null) {throw new TypeError();}for (var key in source) {if (hasOwn.call(source, key) && !hasOwn.call(exclusion, key)) {rest[key] = source[key];}}return rest;})($__0,{bounds:1,map:1,url:1});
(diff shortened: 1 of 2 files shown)


PaulLeCam:
Hey, does it really need to remove and re-create the layer?
Why not calling the element methods instead to handle the changes?

scherler:
Hi Paul, I am not sure whether I understand you. If you refer to the setUrl method then I tried that but it is not getting re-rendered.

PaulLeCam:
Hey, I updated the ImageOverlay component to make the url and opacity properties dynamic, can you please check with v0.3 if that solves your problem?

PaulLeCam:
I'm closing it now due to inactivity, but please feel free to reopen if you need.

scherler:
Hi Paul, sorry for not finding the time to come back earlier on this.

It is working for the src however if I have different dimensions in the image it gets distorted. It maybe a problem with the bounds being different something like the following would fix it I guess if it would exist.

pseudo code:
***************

    if (this.props.bounds !== prevProps.bounds) {
      this.getLeafletElement().setBounds(this.props.bounds);
    }

PaulLeCam:
I see, it seems it can't be handled directly by a Leaflet method on ImageOverlay according to the documentation, but I'll check.
Just out of curiosity, if you set a different `key` prop on the `<ImageOverlay />` component, does it work as expected?

scherler:
Hi Paul, just tried that and yes if I set the key prop than I can use your version as well with different bounds. Thanks.

PaulLeCam:
I'm glad it works, that way React properly treats it as a new image rather than the same, so it gets properly attached.

PR #12, opened 2015-02-13
Add support for event handlers on Map

andywer:
Support attaching DOM event listeners to the <Map /> using React properties.

For instance:

```javascript
<Map onClick={myHandler} />
```

### DIFF
--- a/lib/Map.js
+++ b/lib/Map.js
@@ -66,13 +66,29 @@ var Map = React.createClass({
     this.getLeafletElement().remove();
   },
 
+  getEventHandlers:function() {
+    var handlers = {};
+
+    for (var propKey in this.props) {
+      if (!this.props.hasOwnProperty(propKey)) continue;
+
+      if (propKey.match(/^on/)) {
+        handlers[propKey] = this.props[propKey];
+      }
+    }
+
+    return handlers;
+  },
+
   render:function() {
     var map = this.getLeafletElement();
+    var handlers = this.getEventHandlers();
+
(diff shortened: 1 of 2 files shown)


PaulLeCam:
Hi,
Leaflet event handlers can be attached to the `Map` using the `onLeaflet` prefix (https://github.com/PaulLeCam/react-leaflet#events), but not the DOM ones.
Is there any use case it would be an absolute necessity to bind DOM events?

andywer:
Hi Paul,
we are currently using the handlers onDragOver (just for event.preventDefault()) and onDrop to enable dragging objects from outside into the map. This doesn't work using the Leaflet event handlers (since Leaflet handles d&d in a way that doesn't suite our needs).

Cheers,
Andy

PaulLeCam:
Hi Andy,
Thanks for the context.

At the moment I'm still unsure whether DOM events should be supported or only Leaflet ones.
The Leaflet ones are prefixed with `onLeaflet` so they wouldn't clash with DOM ones in the props, but at the same time my concern is that attaching both DOM and Leaflet events could lead to confusing behaviors.
The only case I've had where DOM events would be useful is on the Map, as in your case, and what I did was simply to wrap the Map in another component to handle these events, do you think this would work for you?

At some point I'd like to simply set the Leaflet events as `on...` rather than  `onLeaflet...`, any suggestion?

andywer:
Yeah, thanks for the hint. The mark-up doesn't look as clean now, but I totally understand your point.
Your idea of changing the listener property names from `onLeaflet...` to `on...` sounds good to me.

Cheers

Issue #11, opened 2015-02-12
Make popup clickable and interactable


scherler:
Hi Paul,

I am trying to use clickable elements in popups, but without success. 

For example imagine the following:

```javascript
    <Marker position={position}>
      <Popup>
          <button onClick={this.bus}>thorsten</button>
        </Popup>
    </Marker>
```

Now if you click on the button there is nothing happening (at least there should be a complain about "bus" not defined. On a side note I tried as well with "onclick" and as well with onClick="this.bus()"

Now when I patch the marker like:

    --- a/src/Marker.js
    +++ b/src/Marker.js
    @@ -16,6 +16,9 @@ module.exports = React.createClass({
       componentWillMount() {
         var {map, position, ...props} = this.props;
    -    this._leafletElement = Leaflet.marker(position, props);
    +    var marker = Leaflet.marker(position, props);
    +    if (this.props.onClick) {
    +      marker.on('click', this.props.onClick);
    +    }
    +   this._leafletElement = marker;
       },

And do

    <Marker position={position}  onClick={this.bus}"/>

Then it is working as expected but I have no popup.

To bring back the popup and not having React.renderToStaticMarkup (which is in the Popup.js) remove all onClicks or logic I use, I came up with the following:

    --- a/src/Marker.js
    +++ b/src/Marker.js
    @@ -16,6 +16,9 @@ module.exports = React.createClass({
       componentWillMount() {
         var {map, position, ...props} = this.props;
    -    this._leafletElement = Leaflet.marker(position, props);
    +    var marker = Leaflet.marker(position, props);
    +    if (this.props.onClick) {
    +      marker.on('click', this.props.onClick);
    +    }
    +   this._leafletElement = marker;
    +    var popup = this.props.popup;
    +    if (popup) {
    +      var LPopup = Leaflet.popup().setContent(popup);
    +      marker.bindPopup(LPopup);
    +    }
    +    this._leafletElement = marker;
       },

and in my calling code I have:

```javascript
   function bus(){
     console.log("ssssssssserererererssss")
   }

   Leaflet.thorsten={};
   Leaflet.thorsten.bus = bus;
...
    var  xxx = '<div><h2>bus</h2><button onclick="L.thorsten.bus()">thorsten</button></div>';
     <Marker position={position} popup={xxx}"/>
```

This finally works, but it really feels like a workaround more then the way to go.

Do you know a better way?

PaulLeCam:
Hi,
[...]

PR #13, opened 2015-02-18
#11 Mixin solution for the issue

scherler:
Creating a Mixin which allows to use React content in leaflet popups

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.idea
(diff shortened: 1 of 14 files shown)


PaulLeCam:
Hi,
I think this kind of code should go in the application, not the library.

At this point I don't want to add specific handlers like `createPopup()` to be able to create a Popup, in my opinion it goes against the point of using React and simply having to include a component to render it.
I'll try to find a solution to render React components in Popups that doesn't involve new APIs.

PR #22, opened 2015-04-27
Use leaflet-headless instead leaflet

### LINKED ISSUE #20: Window is not defined
Hi,

I am trying to use react-leaflet in a isomorphic js project with webpack and leaflet its returning `window is not defined`. Is there any reason to use window object in the server side?

itoldya:
Fix #20

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 coverage
+.idea/
--- a/lib/BaseTileLayer.js
+++ b/lib/BaseTileLayer.js
@@ -1,6 +1,6 @@
 "use strict";
 
-var _interopRequireWildcard = function (obj) { return obj && obj.__esModule ? obj : { "default": obj }; };
+var _interopRequireDefault = function (obj) { return obj && obj.__esModule ? obj : { "default": obj }; };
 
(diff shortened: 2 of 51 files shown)


PaulLeCam:
Hi,

Could you please give more details about how this change affects the library please?
I have never tried leaflet-headless, but from my understanding it renders a static map to an image, which is not the purpose of this library.

itoldya:
[Leaflet](https://github.com/Leaflet/Leaflet) throw an exception `ReferenceError: window is not defined ` when you try to use it in nodejs.
[Leaflet-hedless](https://github.com/jieter/leaflet-headless) solves this problem by using jsdom. Leaflet-headless also can export the map as an image, but we do not need that.

Using Leaflet-hedless instead Leaflet means the following:
- components created with the react-leaflet can be rendered on the server, ie can be used for isomorphic applications
- [nothing will change when using in browser](https://github.com/jieter/leaflet-headless/blob/master/browser.js)

PaulLeCam:
As much as I'd like to support server-side rendering, I don't think it's the way forward. leaflet-headless has too constraining dependencies to be a simple drop-in replacement.

For now, this library is only focused on proxying to Leaflet, not offer additional features that are not supported by Leaflet.

evantahler:
@itoldya did you ever publish your fork of this project using leaflet-headless?  I'm having the same problems you are :/

Issue #18, opened 2015-04-06
Problems with findComponentRoot


Otann:
Hi!

I have some weird error, while using your lib.
I wonder if it is possible to use your components inside old-style react createClass components?

When I use simple component everything seems to work fine:

```jsx
import React from 'react'

// import {Map, Marker, Popup, TileLayer} from "react-leaflet";

module.exports = React.createClass({
  displayName: 'Satellite',

  render: function() {
    const position = [this.state.lat, this.state.lng];
    return (
      <h1>content</h1>
    );
  },

  getInitialState: function() {
    return {
      lat: 51.505,
      lng: -0.09,
      zoom: 13
    };
  }

});
```

But then I add simple map from your example:

```jsx
import React from 'react'

import { Map, Marker, Popup, TileLayer } from 'react-leaflet';

module.exports = React.createClass({
  displayName: 'Satellite',

  render: function() {
    const position = [this.state.lat, this.state.lng];
    return (
      <Map center={position} zoom={this.state.zoom}>
        <TileLayer
          url="http://{s}.tile.osm.org/{z}/{x}/{y}.png"
          attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
        />
        <Marker position={position}>
          <Popup>
            <span>A pretty CSS3 popup. <br/> Easily customizable.</span>
          </Popup>
        </Marker>
      </Map>
    );
  },

  getInitialState: function() {
    const flux = this.getFlux();
    flux.actions.drafts.reloadAsync();
    return {
      lat: 51.505,
      lng: -0.09,
      zoom: 13
    };
  }

});
```

And I have following error in browser:

```
Error: Invariant Violation: findComponentRoot(..., .0.0.0.$=11:0.$=10:0): Unable to find element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID ``.
```

Am I forgetting about something or doing something wrong?

PaulLeCam:
Hi,

Yes it's possible to use it with `React.createClass()`, as long as it's using React 0.13+ the components definitions will be handled by React.

Looking at your code, how does it behave when you remove these lines?
```js
const flux = this.getFlux();
flux.actions.drafts.reloadAsync();
```

PaulLeCam:
[...]

PR #25, opened 2015-05-03
Fix for issue #18

Frikki:
Changing the element from `noscript` to, for example, `div` fixes the `Error: Invariant Violation: findComponentRoot`, which is experienced by both flummox users and raised when the browser navigation’s back and forward is used. See https://github.com/PaulLeCam/react-leaflet/issues/18


PaulLeCam:
Hi, thanks for the fix but it would be hiding the problem rather than solving it, see https://github.com/PaulLeCam/react-leaflet/issues/18#issuecomment-99013617 for a more detailed explanation.

PR #28, opened 2015-05-08
Allow inline styling for Map comp

troutowicz:
For React projects that are starting to adopt the inclusion of CSS in components, this change would be super helpful.

[example](https://github.com/troutowicz/geoshare/blob/css-in-js/app/components/App.jsx#L40)


PaulLeCam:
Good idea, thanks!

PR #32, opened 2015-05-15
If fill colour is changing we need to re-render the form

scherler:


PaulLeCam:
Hi,

At this point I don't want to support individual styling attributes on the elements that do not have specific Leaflet setters for them, this kind of logic should be part of your application code rather than the lib.

PR #36, opened 2015-05-20
Fixed exporting of PopupContainer

adanielyan:
PopupContainer will export PopupContainer instead of Popup.

It actually gave me a hard time to figure out why my own Leaflet elements inherited from PopupContainer did not render. It turned out that the PopupContainer was exporting Popup instead. After fixing this everything worked as expected.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,7 +18,7 @@ import MultiPolyline from "./MultiPolyline";
 import Polygon from "./Polygon";
 import Polyline from "./Polyline";
 import Popup from "./Popup";
-import PopupContainer from "./Popup";
+import PopupContainer from "./PopupContainer";
 import Rectangle from "./Rectangle";
 import TileLayer from "./TileLayer";
 import WMSTileLayer from "./WMSTileLayer";


PaulLeCam:
Thanks!

hokkos:
Oh, finally, I extended PopupContainer and couldn't understand why my custom layer was never added to the map: the function componentDidMount was never called. Please push a new version to npm I can't even use the one from github. I had to modify the package by hand.

adanielyan:
To use the repo from Github add this line to your package.json file under dependences:
```
"react-leaflet": "git+https://github.com/PaulLeCam/react-leaflet.git"
```

troutowicz:
@adanielyan Or the much shorter way...

`"react-leaflet": "PaulLeCam/react-leaflet"`

adanielyan:
Thanks for the shortcut, I didn't know about it!

adanielyan:
By the way, the code in lib folder doesn't have this change. Make sure you recompile the repo after pulling it. It was another half an hour of struggling for me.

hokkos:
Using the git repository was the first thing I tried, but I don't know why it only download a package with a lib/ directory inside, not the git one with src/ inside.

PR #41, opened 2015-06-11
Fix ignored events when there is an attribute named "length" in a layer

amarant:
It fixes an issue when a layer component had a length attribute, lodash see the object as an array so the events couldn't be extracted.

Also fix nested simple quote issues after a blind double to simple quote replace.
I didn't included the generated files.


PaulLeCam:
Thanks!

PR #50, opened 2015-07-21
Add bower.json file

lennerd:
I use this library in conjunction with the [Rails Assets proxy](https://rails-assets.org/) to use bower dependencies in my gem file. Unfortunately this library does not provide a bower.json file and is therefore not part of the bower registry.

This PR adds a `bower.json` file. The registration must be done by the author.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
+bower_components
 coverage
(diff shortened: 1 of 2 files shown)


PaulLeCam:
Thanks!

PR #55, opened 2015-07-31
Single file build script

### LINKED ISSUE #29: Standalone build?
Hello!
I would really love to use react-leaflet in my project but i don't have a require.js there and i can't have it installed for various reasons. I just can't. 

I can't see any standalone build of react-leaflet in the repo but i've seen a lot or react (and js in generall) libs made in two version - for AMD and standalone so i assume that's doable. I'm not really good in JS and i don't know require.js so i'm totally not sure if that's possible but - is that possible for you to add a standalone version ? Or can i somehow build such thing myself?

Thanks!

gambit3:
This is a custom build script I use to build a single (but unminified) version of react-leaflet. It should fix #29


package.json
```diff
@@ -13,7 +13,8 @@
     "dev": "npm run watch & npm run onchange",
     "prepublish": "npm ls && npm test",
     "example:compile": "gulp",
-    "example:watch": "gulp watch:example"
+    "example:watch": "gulp watch:example",
+    "build_single": "node single_build_globals.js"
```
luqmaan: For consistency should this be `build:single`?

single_build_globals.js
```diff
@@ -0,0 +1,31 @@
+var browserify = require('browserify');
+var babelify = require("babelify");
```
luqmaan: Use `'` vs `"` consistently.

package.json
gambit3: Sure. The script should probably also live somewhere else and always build the same file. I didn't know how to organise it to be honest, just wanted to share the build script.

PaulLeCam:
Thanks but I plan to move over to Webpack for the build, I'll make it export to UMD.

PR #56, opened 2015-07-31
Fix latLngList not supporting polygons with holes

gambit3:
see http://leafletjs.com/reference.html#polygon
When creating polygons with holes react-leflet complains is has wrong params.
Array of arrays of latlngs should also be allowed since Leaflet supports them.


luqmaan:
:+1:

PaulLeCam:
Fixed in v0.7.

PR #57, opened 2015-07-31
Optimise marker and polygon componentDidUpdate

gambit3:
This is an optimisation I use so that if a marker or a polygon are re-rendered often the DOM is not touched if their props didn't change. 
The previous approach only works with immutable latLng and positions objects and it's sometimes difficult to achieve that. This fixes it a little.

### DIFF
--- a/src/Polygon.js
+++ b/src/Polygon.js
@@ -1,7 +1,10 @@
 import Leaflet from 'leaflet';
+import isEqual from 'lodash/lang/isEqual';
+import find from 'lodash/collection/find';
 
 import latlngListType from './types/latlngList';
 import PopupContainer from './PopupContainer';
+
 
 export default class Polygon extends PopupContainer {
   componentWillMount() {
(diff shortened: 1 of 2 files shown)


src/Polygon.js
```diff
@@ -12,7 +15,14 @@ export default class Polygon extends PopupContainer {
 
   componentDidUpdate(prevProps) {
     if (this.props.positions !== prevProps.positions) {
-      this.leafletElement.setLatLngs(this.props.positions);
+      if(!isEqual(this.props.positions, prevProps.positions)){
```
luqmaan: Why do you need to use `isEqual` in addition to the `!==` operator?

gambit3: Well that was the whole point. Say in a render method you'll do something like this
``` javascript
var positions = _.map(this.props.geoJsonPoints, (point)->{  return [point[1], point[0]})
// or some other transformation of props that result in a new positions array
render <Polygon positions={positions} />
```
than you're constructing a new positions array on each render and `!==` does not wor. You are re-rendering the polygon even its positions didn't change (leaflet doesn't check that internally either) ie. touching the DOM which is bad for performance and very bad if you are re-rendering often especially polygons.
You can obviously work around that in your render method but if you're not aware of it its easy to miss. So this is just a react'y way of not touching the DOM if we don't have to.

PaulLeCam:
I think this kind of optimisation should be part of the application, not the base components like this.

PR #58, opened 2015-07-31
New component: LayerGroup 

gambit3:
It can wrap other layers inside itself. (http://leafletjs.com/reference.html#layergroup)
I used this a as basis for MarkerClusterGroup to enable clustering capabilities using leaflet-markercluster plugin. (see my other PR)
I also think the FeatureGroup component should extend this one so it can actually wrap other react-leflet components not just regular Leaflet objects

### DIFF
--- a/src/LayerGroup.js
+++ b/src/LayerGroup.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import Leaflet from 'leaflet';
+import MapLayer from './MapLayer';
+import { layerGroup } from 'leaflet';
+
+export default class LayerGroup extends MapLayer {
+  componentDidMount() {
+    super.componentDidMount();
+    (this.props.layerGroup || this.props.map).addLayer(this.leafletElement);
+  }
+
+  componentWillMount() {
+    super.componentWillMount();
+    const { map, ...props } = this.props;
+    this.leafletElement = layerGroup();
+  }
+
+  render() {
+    const layerGroup = this.leafletElement;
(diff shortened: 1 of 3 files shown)


src/LayerGroup.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import Leaflet from 'leaflet';
+import MapLayer from './MapLayer';
+import { layerGroup } from 'leaflet';
```
luqmaan: Why switch to this syntax?

gambit3:
I've notice @PaulLeCam switch do this here:
https://github.com/PaulLeCam/react-leaflet/commit/9945c662b933b6e817ac1e74fe8357d0fde1731f#diff-873cfcc6333674fc88cf194c7eae1641L1
and assumed this was a convention of some sort he adopted

src/LayerGroup.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default class LayerGroup extends MapLayer {
+  componentDidMount() {
```
PaulLeCam: This handler seems extraneous considering it's set in `MapLayer`, no?

PaulLeCam:
Hi,
Thanks for the PR, could you please add tests to check the behavior when (un)mounting a `LayerGroup` on a `Map` and other layers on a `LayerGroup`?

PaulLeCam:
Added in v0.8.0-rc.3, thanks!

PR #59, opened 2015-07-31
New component: MarkerClusterGroup 

gambit3:
Extends LayerGroup and provides clustering functionality for markers via external plugin (a new peerDependency): leaflet.markercluster.


luqmaan:
This shouldn't be part of the react-leaflet library. It should live outside. For example see https://github.com/paulserraino/react-geojson-cluster

PR #60, opened 2015-07-31
Popup content working using react 

gambit3:
Replaced react.renderToStaticMarkup with proper react.render to allow for dom diffing and dynamic updates on popup content.
There is still a single child limitation ie. <Popup> can only have one child, but apart from that anything placed inside the popup will behave like normal react.dom content with proper diffing and updating when popup is opened.
It works by creating a new react root each time a popup is opened and scrubbing it if it closed

### DIFF
--- a/src/Popup.js
+++ b/src/Popup.js
@@ -11,45 +11,54 @@ export default class Popup extends MapComponent {
 
   componentWillMount() {
     super.componentWillMount();
-    const { children, map, ...props } = this.props;
-    this.leafletElement = popup(props);
+    const { children, map, popupContainer, ...props } = this.props;
+    this.leafletElement = popup(props, popupContainer);
+    this.leafletElement.on('open', () => { //on popoup create new react root
+      const {children} = this.props
+      this.contentComponent = React.render(React.DOM.div({children}), this.leafletElement._contentNode)
+      this.leafletElement._adjustPan()
+    }, this)
+    this.leafletElement.on('close', () => { //remove react root
+      delete this.contentComponent
+      React.unmountComponentAtNode(this.leafletElement._contentNode)
+    }, this)
+  }
+
+  componentDidMount() {
+    const { popupContainer, map, position } = this.props;
+    super.componentDidMount();
+    //Attach to container component
+    if (popupContainer) {
+      popupContainer.bindPopup(this.leafletElement)
+    } else { //attach to map
+      if(position) this.leafletElement.setLatLng(position)
+      this.leafletElement.openOn(map)
+    }
   }
 
   componentDidUpdate(prevProps) {
-    const { children, map, popupContainer, position, ...props } = this.props;
-    if (children !== prevProps.children) {
-      const content = React.renderToStaticMarkup(children);
-      if (popupContainer) {
-        popupContainer.bindPopup(content, props);
-      }
-      else {
-        const el = this.leafletElement;
-        el.setContent(content);
-        if (position !== prevProps.position) el.setLatLng(position);
-      }
+    const { children, popupContainer, position } = this.props;
+    if(!popupContainer) {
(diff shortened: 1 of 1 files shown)


src/Popup.js
```diff
@@ -11,45 +11,54 @@ export default class Popup extends MapComponent {
...
-      if (popupContainer) {
-        popupContainer.bindPopup(content, props);
-      }
-      else {
-        const el = this.leafletElement;
-        el.setContent(content);
-        if (position !== prevProps.position) el.setLatLng(position);
-      }
+    const { children, popupContainer, position } = this.props;
+    if(!popupContainer) {
```
luqmaan: This should be:

```js
if (!popupContainer) {
```

Or at least use consistent style. You switch between `if(` and `if (` throughout this file. :smiling_imp:

src/Popup.js
```diff
@@ -11,45 +11,54 @@ export default class Popup extends MapComponent {
...
+    const { map, popupContainer } = this.props;
     super.componentWillUnmount();
-    this.props.map.removeLayer(this.leafletElement);
+    if(this.leafletElement._contentNode) {
+      React.unmountComponentAtNode(this.leafletElement._contentNode);
+    }
+    if(popupContainer) {
+      popupContainer.unbindPopup(this.leafletElement);
+    } else {
+      map.removeLayer(this.leafletElement)
```
luqmaan: Missing semicolon.

src/Popup.js
```diff
@@ -11,45 +11,54 @@ export default class Popup extends MapComponent {
...
     super.componentWillMount();
-    const { children, map, ...props } = this.props;
-    this.leafletElement = popup(props);
+    const { children, map, popupContainer, ...props } = this.props;
+    this.leafletElement = popup(props, popupContainer);
+    this.leafletElement.on('open', () => { //on popoup create new react root
+      const {children} = this.props
+      this.contentComponent = React.render(React.DOM.div({children}), this.leafletElement._contentNode)
+      this.leafletElement._adjustPan()
+    }, this)
```
luqmaan: You don't need to specify the `this` argument since you're using a fat arrow.

luqmaan:
Great idea

src/Popup.js
gambit3: Just a habit. I'm not used to es6 yet this is my first time really using it.

src/Popup.js
gambit3: Sure. I use different style when I write my other code so its easy to miss for me.

src/Popup.js
gambit3: Again I'm one of those 'heretics' who don't use semicolons when writing my own stuff, so it's easy for me to miss this stuff. Do you perhaps have some recommendations for a linter (which plugs to sublime2 ideally) that would work on es6 and point out those kind of style stuff for me?

bohlander:
This looks really handy -- any tips on the best way to incorporate this change into my npm installed version of react-leaflet? I've tried tweaking my package.json to refer to the branch containing the pull request, but I don't think this is a good approach, seems like it needs to be explicitly built somehow.

gambit3:
@bohlander I don't know exactly how to hack npm like that since I use a standalone build from /src. If you tweaked package.json my guess is you'd have to go to the` node_modules/react-leaflet` dir and run `npm compile` to rebuild the `/lib` files inside react-leaflet package. I haven't included `/lib/*` files in this pr since it would make merging it difficult.

bohlander:
Thanks @gambit3 -- I think I might try using your build script and see if I can produce a standalone version of react-leaflet to try instead. Looking forward to these pull requests getting merged!

PaulLeCam:
Hi, thanks for the PR, I changed a few things but added this logic in v0.8.

PR #72, opened 2015-09-22
Add a BoundsMap (<Map> alternative) component

uniphil:
This pull request adds a component which sets the map view via a dynamic `bounds` prop instead of by `center` and `zoom`.

An alternative implementation could add the `bounds` prop to the existing `Map` component, and let users to either set `bounds` or `center`+`zoom` to control the map view. I started writing it that way, but found it required too many if/else blocks and helper functions to do the right thing for my taste. If you would prefer it that way, I can change it back.

At work we use `map.fitBounds` almost exclusively for setting the view, so something like this component is essential for our use of `react-leaflet`.


src/BoundsMap.js
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  componentDidMount() {
+    super.componentDidMount();
+    if (!isUndefined(this.props.bounds)) {
+      this.leafletElement.fitBounds(this.props.bounds);
+    }
+  }
+
+  componentDidUpdate({ bounds: oldBounds }) {
```
luqmaan: I think this would be better as `componentDidUpdate(props)`, as it makes things more clear to anybody that wants to extend `BoundsMap`.

uniphil: :+1:

PaulLeCam:
Hi, thanks for the PR.

Is there any particular reason for doing it as a separate component rather than adding this behavior to `Map`?
I suppose the `fitBounds()` call could be made after the center and zoom are set, would that suit your use case?

uniphil:
Calling `fitBounds` manually works, but it moves all the bookkeeping logic into our app, which we are trying to avoid with React. The use-case here: We have "international", "country", and "project" views for geocoded data in Asia. We use react-router to draw chrome around the map and features on the map depending on the view. It's nice to just update `bounds` like the rest of our data and have react/react-leaflet deal with it.

If you would prefer, I can modify `Map` to handle a `bounds` prop instead of adding this `BoundsMap` component. In that case:

- Setting `bounds` AND (`center` OR `zoom`) props is probably a mistake by the user
    - it would be nice to log a warning. I don't see any warnings currently used, and `console` statements make the linter complain. Do you have a preferred way to do this? Can I add a `warn.js` module with `/* eslint no-console: 0 */` that exports a `warn` function?
    - If this case happens, which props should react-leaflet use to set the view -- `bounds`, or (`center` and `zoom`)?

PaulLeCam:
I think it makes sense to have the `Map` component handle dynamic `bounds` and `maxBounds` props, it could then be used either using `bounds` or the `center`/`zoom` combination.

I'm against adding any specific logic in the components regarding specific use cases as in my option the components provided in this lib should always be "dumb" ones, the "smart" ones being in the app.
Regarding the case of setting `bounds`, `center` and `zoom`, I think the `Map` component would just apply all of them, it's up to the developer to use the APIs they need.

If you're up for it, would you mind updating your PR to add this to `Map`? Otherwise I'll try to experiment around it when I have some time.

uniphil:
Moved it all to dynamic `bounds` on `<Map>`: no validating or branching logic based on props -- last update wins.

PR #78, opened 2015-09-30
Add CartoDBTileLayer component

ericsoco:
I'm not sure it's a great thing to bring this into `react-leaflet` until CartoDB publishes their core API (with the necessary `Tiles` dependency) to npm. But, wanted to get this to you for a first look.

Any thoughts? It's rough, could use some documentation, but it works and is useful for us.

### DIFF
--- a/example/build/dependencies.js
+++ b/example/build/dependencies.js
@@ -1,1227 +1,4 @@
 require=(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
-(function (process){
-/**
- * Copyright 2013-2015, Facebook, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


ericsoco:
BTW, I noticed the `lint` step fails on `Map` and `Popup`. The content of this PR should be clean.

PaulLeCam:
Hi,
I fixed the lint errors, thanks for reporting.

Regarding adding this kind of features, I don't want to add it to the "core" library as it would add dependencies that would not necessarily be needed.
The way I see it, there are 3 major options:

1. Adding these features to the lib, not exposed in the API (importing `react-leaflet` module), but rather having to import from another directory (ex importing `react-leaflet/plugins/CartoDBTileLayer`).
  - Advantages are that plugins would be updated with the lib, so it's easier for consumers to update. It's also easy to keep track of the plugins and to add new ones.
  - Disadvantages are that it would require more work to maintain this lib, especially to update it with newer versions of React and Leaflet, as the plugins dependencies would also need to support these new versions.

2. Creating a separate repository for plugins, ex `react-leaflet-plugins` having `react-leaflet` as a dependency, and same as previous option, each module would have to be loaded individually so that only the dependencies for this module would be required.
  - Advantages is that it keeps this lib only dependant on React and Leaflet, so easier to update, while having most of the advantages of including the plugins with the lib.
  - Same disadvantages as integrating with the lib, one plugin dependency could hold back updating others.

3. All plugins should be separate projects.
  - Advantages are that this lib and the plugins are loosely coupled, they can be updated completely independently. It also leaves full control with plugins authors rather than having to PR this lib.
  - Disadvantages I can think of would be plugins discovery, but it's not a major issue thanks to npm search, and the documentation in this lib could always link to the plugins.

My preference really goes for the third option, the disadvantages of the other two are much bigger than the advantages in my option.
Any thought?

ericsoco:
@PaulLeCam agreed. I'll report back when I've gotten this up on its own repo, and maybe we can add a link to it from your documentation in a shiny new **Plugins** section. Thanks!

PaulLeCam:
Great, thanks!

PR #82, opened 2015-10-21
fix the export of PropTypes

amarant:
The PropTypes weren't properly exported, because in /types/index.js there is no default export, and PropTypes was the default name, so now we import all then re-export.


PaulLeCam:
Hi,
Thanks for reporting this, I pushed v0.8.1 that fixes this.

PR #89, opened 2015-10-29
Control open/close state of popup

stevenmason:
Added open bool prop to popup which allows control of open/close state
of a popup


PaulLeCam:
I don't think this can work, Leaflet maintains its own state for the popup and as far as I can tell there is no way to prevent it from opening/closing a popup from the events, so the value of the prop would not match the actual popup state.

stevenmason:
Good point, I needed this functionality for the application I am making, the only reason it seems to work for me is that I set the props open state in the obvious cases where leaflet changes the state for example on click of a marker and clicking on the map.

PR #96, opened 2015-11-09
Use unstable_renderSubtreeIntoContainer

dallonf:
This ensures that `context` from the parent context bubbles down to the Popup. Most React libraries, especially for data management and localization, rely on `context` pretty heavily. Despite the `unstable_` prefix, this is pretty much the most elegant way to do this at the moment.


PaulLeCam:
Hi,

Thanks for your PR but I don't want to use undocumented React APIs that could be removed at any time without warning.
I hope this function would be added to the official API in a next release!

Issue #87, opened 2015-10-29
v1 todo


PaulLeCam:
## Must have

- [x] UMD build ([#29](https://github.com/PaulLeCam/react-leaflet/issues/29))
- [x] Support for Controls ([#81](https://github.com/PaulLeCam/react-leaflet/issues/81))
  - [x] Zoom
  - [x] Attribution
  - [x] Layers
  - [x] Scale
- [x] Contributing guidelines
  - [x] Setup jsbin/jsfiddle with dependencies to reproduce bugs (http://jsfiddle.net/paul_lecam/q2v7t59h/)
  - [x] Add `CONTRIBUTING.md`
  - [x] Add `ISSUE_TEMPLATE.md`
- [x] Remove support for `onLeaflet...` events
  - [x] Add deprecation warning in previous release
- [x] Extract docs out of readme, split in different sections
- [x] Remove support for `getLeafletElement()` method
  - [x] Add deprecation warning in previous release
- [x] Support only React v15+
- [ ] Leaflet v1 support ([#66](https://github.com/PaulLeCam/react-leaflet/issues/66))

## Nice to have

- [ ] Website
  - [ ] Getting started
  - [ ] API
  - [ ] Examples
- [ ] Examples for each component
- [ ] Advanced examples (Redux, Rx...)
- [ ] Tutorial for creating custom components
  - [ ] Reference implementation for plugin integration

jonboiser:
Hi Paul,

I like the potential of this library for quickly bootstrapping geospatial apps written in React. Are there are any tasks you can use help with?

PaulLeCam:
Hi @jonboiser, thanks for asking!

I think adding the attribution, layers and scale controls is the next priority. It shouldn't be a lot of work, just make sure they behave as expected.

Leaflet v1 support is tricky because I want to have it as soon as it's stable, but the current betas seem to change the APIs so I'm not sure it's worth investing time on it.

jonboiser:
Cool. I'll take a look at how you've implemented the Zoom control. I think I can help with the docs and website, once I play around with the library. At my company, we started a similar project for the ArcGIS JS API and Leaflet, but think we should deprecate the Leaflet piece, since this library is available.

jonboiser:
I pushed some code for the AttributionControl to my fork. It seems like coding the other controls will be pretty much the same. I'll tackle the others during the week and submit a PR.

snario:
What's the progress like on this?

[...]

PR #100, opened 2015-11-15
Add other controls in v1 TODO

jonboiser:
Filling out some of the missing controls in #87.

I've done some basic testing on Attribution and Scale. The Layers control gets placed on the map, but I'm not sure how to correctly pass a base layer reference from a, say, `TileLayer` component into the Layers control.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,19 @@ Extended `LayerGroup` supporting a `Popup` child.
 
 #### Controls
 
+##### AttributionControl
+
+- `position: controlPosition` (optional, dynamic)
+- `prefix: string` (optional)
+
+##### ScaleControl
+
+- `imperial: bool` (optional)
+- `position: controlPosition` (optional, dynamic)
+- `maxWidth: number` (optional)
+- `metric: bool` (optional)
+- `updateWhenIdle: bool` (optional)
+
 ##### ZoomControl
 
 - `position: controlPosition` (optional, dynamic)
--- a/src/AttributionControl.js
+++ b/src/AttributionControl.js
@@ -1,0 +1,14 @@
+import { PropTypes } from 'react';
+import { control } from 'leaflet';
+
+import MapControl from './MapControl';
+
+export default class AttributionControl extends MapControl {
+  static propTypes = {
+    prefix: PropTypes.string,
+  };
+
+  componentWillMount() {
+    this.leafletElement = control.attribution(this.props);
+  }
+}
--- a/src/LayersControl.js
+++ b/src/LayersControl.js
@@ -1,0 +1,17 @@
+import { PropTypes } from 'react';
+import { control } from 'leaflet';
+
+import MapControl from './MapControl';
+
+export default class LayersControl extends MapControl {
+  static propTypes = {
+    baseLayers: PropTypes.object,
+    options: PropTypes.object,
+    overlays: PropTypes.object,
+  };
+
+  componentWillMount() {
+    const { baseLayers, overlays, options } = this.props;
+    this.leafletElement = control.layers(baseLayers, overlays, options);
+  }
+}
--- a/src/ScaleControl.js
+++ b/src/ScaleControl.js
@@ -1,0 +1,17 @@
+import { PropTypes } from 'react';
+import { control } from 'leaflet';
+
+import MapControl from './MapControl';
+
+export default class ZoomControl extends MapControl {
+  static propTypes = {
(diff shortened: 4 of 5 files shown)


jonboiser:
Here is one solution I've been working on for the LayersControl. In your `Map`, you can include multiple layers as children.

```{jsx}
<Map bounds={this.state.bounds}>
  <TileLayer
    attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
    layerName='OpenStreetMap'
    url='http://{s}.tile.osm.org/{z}/{x}/{y}.png'
  />
  <TileLayer
    attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
    layerName='OpenStreetMap2'
    url='http://otile4.mqcdn.com/tiles/1.0.0/osm/{z}/{x}/{y}.png'
  />
  <LayersControl baseLayers={['OpenStreetMap', 'OpenStreetMap2']}/>
</Map>
```

The difference is that you pass `layerName` props into the layers, and pass an array matching those names as a prop into `LayersControl`.

I'm not sure we can look at all the layers synchronously, so in the `LayersControl` `componentWillMount` stage, we listen for these layers being added to the map asynchronously. When the layer is added, we also add it to the control. When we are sure the last layer is added, we remove the event listener.

PaulLeCam:
Hey, thanks for the PR!

I'll try to have a look at Leaflet's doc this week for these controls, see if there any specific cases to consider.
Other than that I think it will mostly be about handling dynamic properties and adding some doc, but your current code looks good.

PaulLeCam:
Thanks!

chriswiggins:
Sorry to dig this up - I'm having issues with understanding _how_ to use the layers control? It seems to _work_ when I pass actual leaflet objects in as you would normally, but how do you set a default?

Much appreciated - happy to submit a PR with updated Readme instructions and even an example once I get it working :-)

PaulLeCam:
What do you mean by default? The default options are Leaflet's ones, this library simply passes the props to the relevant Leaflet constructor.

chriswiggins:
Hi,

What do you mean by default? The default options are Leaflet's ones, this library simply passes the props to the relevant Leaflet constructor.

OK right I see. I was wondering if it was possible to pass in a <TileLayer> through to LayersControl but it looks like as you say it's just a shim around the leaflet options. In the leaflet options for the L.control.layers you need to pass an object where the key is the title shown on the control and the value is a L.TileLayer class. I was hoping instead I could pass a <TileLayer> from here :-)

Thanks for the help

TobyEalden:
Hi - I'm having a similar issue - I have a `<GeoJson>` overlay as a child of `<Map>` - how do I configure the `<LayersControl>` `overlays` property to reference the `<GeoJson>` element? Do I need to write some code in componentDidMount to discover the Leaflet element corresponding to the GeoJson layer and manually set the overlays property?

jonboiser:
@chriswiggins Could you give a code example of this?

> I was wondering if it was possible to pass in a <TileLayer> through to LayersControl

I think I had a similar idea when I first wrote these wrappers (see my earlier comment), but I think it would involve writing some more logic into the guts of LayersControl to pull it off.

chriswiggins:
@jonboiser I was thinking of something like this:

```
<LayersControl ...layersOpts>
    <TileLayer type='base' title='Street View' selected='true' ...otherOpts/>
    <TileLayer type='base' title='Satellite View' ...otherOpts />
    <ImageOverlay type='Overlay' title='Overlay' ...otherOpts />
</LayersControl>
```

Come to think of it, that could actually be fairly straightforward if the LayersControl element "rendered" its children and then added to the layers control. I'll have a play today and get back to you all on whether or not this works!

chriswiggins:
So I've given this a go and no immediate luck - mainly because the TileLayers are expecting a `map` prop to be passed to them however in this implementation thats not what we want to do. I might have to end up modifying lots of components to get this to work. 

Any other ideas around how we could do this?

jonboiser:
So you would wrap the LayersControl-composed-with-TileLayers in a Map like this? I think this would be a nice way to work with Leaflet.
```
<Map>
<LayersControl ...layersOpts>
    <TileLayer type='base' title='Street View' selected='true' ...otherOpts/>
    <TileLayer type='base' title='Satellite View' ...otherOpts />
    <ImageOverlay type='Overlay' title='Overlay' ...otherOpts />
</LayersControl>
</Map>
```

PR #110, opened 2015-12-20
Feature/edit control

alex3165:


PaulLeCam:
Hi,

Thanks for your PR but this project intends to be a simple interface for Leaflet, not implement its plugins, please read [this comment](https://github.com/PaulLeCam/react-leaflet/pull/78#issuecomment-144663021) for a more detailed explanation.

As stated in this previous comment, if you want to add this plugin as a separate project, I'd be glad to reference to it from the documentation here.

alex3165:
Hi Paul,

Ok I understand sorry i didn't see the comment before, i will create a
separate project then :) . I have another suggestion, could we add
"react-pure-render" to the different layers so that every time we update
the main component we don't necessarily update all the react-layers ?

Thanks,
Alex.

2015-12-21 10:06 GMT+00:00 Paul Le Cam <notifications@github.com>:

PR #112, opened 2016-01-20
fix the popup source property

amarant:
it cannot call the onpopupopen event handler if it is not set

Try to click the blue outer circle from the example other-layers.js, without the modification the increment function of onpopupopen is not called.


PaulLeCam:
Hi,
Thanks for the fix, it will go in the next release.

PR #117, opened 2016-02-02
Leaflet 1.0

PatrickKing:
Brings the 1.0 branch up to date with master, brings in Leaflet 1.0.0-beta.2, make some small changes to Popups to work with Leaflet 1.0. 

Some tests broken, due to Leaflet 1.0 using clientHeight/clientWidth and these attributes not being emulated in JSDom.


PaulLeCam:
Hi,

Thanks for your PR. I haven't looked more at the changes in Leaflet v1 as it's still not out of beta, but it seems your PR breaks the tests, would you mind having a look at it?

PatrickKing:
The failing tests are due to changes in Leaflet, it now reads clientHeight and clientWidth, which are properties that JSDom does not emulate. The result of the missing numbers is that NaNs are stored for some internal bookkeeping, this gives us the three `Invalid LatLng object: (NaN, NaN)` errors. 

I'm not sure how to fix this problem short of replacing JSDom and reworking the test suite.

PaulLeCam:
I see, it's a tricky issue.
I don't think replacing jsdom is really an option considering it's part of Jest, but maybe by mocking the relevant getter in Leaflet to return an acceptable value?

PatrickKing:
After looking into it: I see that Leaflet is not automocked by Jest. I added this to each of the test files (Popup.js and Marker.js) that was seeing failures, and all tests are passing ... 

`// Mock a Map function that doesn't work without a real DOM
import Leaflet from 'leaflet'
Leaflet.Map.prototype.getSize = () => {
  return new L.Point(1024, 768);
}
`
It works, but is this a good approach?

PaulLeCam:
No, Leaflet is not mocked, so I think only mocking the failing method as you suggest is the best option given the constraints. Would you mind updating your PR with this change please?

PatrickKing:
PR updated, sorry for the wait!

PaulLeCam:
Great, thanks!

PR #122, opened 2016-02-19
Added snippet about zoomControl prop

chung-leong:


PaulLeCam:
Hi,
Thanks for the PR!

I realised the documentation may not be explicit enough about properties like this that are simply passed to Leaflet, I added links to the relevant Leaflet documentation sections.

PR #124, opened 2016-03-09
Move css info to install

benzen:


pke:
More importantly the information that the map container also needs to have a height set.

PaulLeCam:
Thanks but this is part of Leaflet documentation, it's out of the scope of this lib.

PR #126, opened 2016-03-10
Animate panning and zooming

mariusandra:
Hey,

I'm in need of a feature where the map pans smoothly (animate=true) when changing the center.

I have made such modifications in this PR, and added an example.

Is this enough to be merged? Anything else I should do?

Thank you!

Marius


PaulLeCam:
Thanks!

mariusandra:
Thank you for merging! Could you also release a new (patch?) version with this change?

PaulLeCam:
Yes I'll publish a new version this week-end.

PaulLeCam:
@mariusandra I just published v0.10.2 containing you commit, thanks!

PR #131, opened 2016-03-31
Lint 'object destructuring filter' consistently

boromisp:
The pattern `const { map, url, ...props } = this.props` usually
failed during linting, because the ```map``` variable is defined but
not used (`no-unused-vars` rule).

After adding the `"varsIgnorePattern": "^_"` option to the
`no-unused-vars` rule, unused properies can be marked as:
`const { map: _, url, ...props } = this.props`.

I also updated the eslint and babel-eslint dependencies, because of
some relevant bugfixes.


PaulLeCam:
Good idea, thanks!

PR #132, opened 2016-04-01
Layer container interface

boromisp:
**Edit (2016-04-02):**
Split the LayersControl code into a new branch, rebased to d057013, removed the built files, and filterd out the `layerContainer` prop everywhere from the layer options objects (same as the `map` prop).

~~This is a two-part PR.~~

**The first part is the `layerContainer` prop.**

Until now a `MapLayer` could receive either a `map` or a `layerGroup` prop. With this PR a `MapLayer` always receives both the `map` and the `layerContainer` props. The `map` is always the same, and the `layerContainer` is either the `map` or the old `layerGroup`.
This way all the map layers have access to the root `Map` instance, and other kind of layer containers can be introduced less intrusively.

~~**The second is a LayersControl implementation for #87** that makes use of the `layerContainer` prop.~~

~~This uses a HoC for the child layers, and acts as a layers container.~~
~~I haven't tested it thoroughly, but for my use case it seems to be working.~~
~~The readme includes an example.~~

### DIFF
--- a/src/CanvasTileLayer.js
+++ b/src/CanvasTileLayer.js
@@ -5,7 +5,7 @@ import BaseTileLayer from './BaseTileLayer';
 export default class CanvasTileLayer extends BaseTileLayer {
   componentWillMount() {
     super.componentWillMount();
-    const { map: _, ...props } = this.props;
+    const { map: _map, layerContainer: _lc, ...props } = this.props;
     this.leafletElement = tileLayer.canvas(props);
   }
 }
--- a/src/Circle.js
+++ b/src/Circle.js
@@ -12,7 +12,7 @@ export default class Circle extends Path {
 
   componentWillMount() {
     super.componentWillMount();
-    const { center, map: _, radius, ...props } = this.props;
+    const { center, map: _map, layerContainer: _lc, radius, ...props } = this.props;
     this.leafletElement = circle(center, radius, props);
   }
 
--- a/src/CircleMarker.js
+++ b/src/CircleMarker.js
@@ -12,7 +12,7 @@ export default class CircleMarker extends Path {
 
   componentWillMount() {
     super.componentWillMount();
-    const { center, map: _, ...props } = this.props;
+    const { center, map: _map, layerContainer: _lc, ...props } = this.props;
     this.leafletElement = circleMarker(center, props);
   }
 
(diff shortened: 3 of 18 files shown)


boromisp:
I'm not really happy with this yet, I'd much rather have a more declarative approach. Something like:


```
class MyMap extends Component {
    state = {
        selectedLayers: new Set(['tile-layer'])
    };

    handleLayersSelected = selectedLayers => this.setState({selectedLayers});

    render() {
        const { selectedLayers } = this.state;
        const { handleLayersSelected } = this;
        return (
             <Map>
                <LayersControl
                    baseLayers={{
                        'tile-layer': 'Tile layer label',
                        'some-other-layer': 'Other layer label'
                    }}
                    overlays={{
                        'optional-marker-layer': 'Markers...'
                    }}
                    selected={selectedLayers}
                    onChange={handleLayersSelected}
                />{selectedLayers.has('tile-layer') ?
                <TileLayer url={https://...} /> : null}{selectedLayers.has('some-other-layer') ?
                <OtherBaseLayer ... /> : null}{selectedLayers.has('optional-marker-layer') ?
                <LayerGroup>
                    {/* markers... */}
                </LayerGroup> : null}
               <... />
            </Map>
        );
    }
}
```

This would require a custom LayersControl widget.

PaulLeCam:
Hi, thanks for this PR!

I like the "layer container" approach, would you mind adding this change only as a separate PR that could be merged before this one?

Regarding the layers control, I agree it would be nice to handle it directly as components, but I'm not sure the `ControlledLayer` HOC is the best way to go, as from my understanding of your code it would shadow the `title` and `type` props that could be used by the wrapped component.
I'll experiment a bit around this to try to come up with possible alternatives that could match both React and Leaflet APIs.

boromisp:
Hi, I updated the PR.

The `ControlledLayer` is the glue, that presents the `LayersController` as a kind of `layerContainer` to the `MapLayer`-s, and sends `title`, `type` and `defaultAdded` props to the `LayersController` for each `MapLayer`.

I haven't looked into it yet, but the leaflet 1.0 might bring some new concepts relevant to this, like custom *map panes* for layer ordering.

https://github.com/Leaflet/Leaflet/issues/1742
http://leafletjs.com/examples/map-panes.html

PaulLeCam:
Thanks!

PR #135, opened 2016-04-05
Layers control

PaulLeCam:
Hi @boromisp,

I updated the `LayersControl` component in this PR to support adding base layers and overlays as children.
I think this solution is simpler than having controlled layers, as it allows to directly use the existing layers, that only need to be contained in either a `LayersControl.BaseLayer` or `LayersOverlay` component, as shown in this example:

```jsx
<LayersControl position='topright'>
  <LayersControl.BaseLayer name='OpenStreetMap.BlackAndWhite'>
    <TileLayer
      attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
      url='http://{s}.tiles.wmflabs.org/bw-mapnik/{z}/{x}/{y}.png'
    />
  </LayersControl.BaseLayer>
  <LayersControl.BaseLayer name='OpenStreetMap.Mapnik'>
    <TileLayer
      attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
      url='http://{s}.tile.osm.org/{z}/{x}/{y}.png'
    />
  </LayersControl.BaseLayer>
  <LayersControl.Overlay name='Marker with popup'>
    <Marker position={[51.51, -0.06]}>
      <Popup>
        <span>A pretty CSS3 popup. <br/> Easily customizable.</span>
      </Popup>
    </Marker>
  </LayersControl.Overlay>
  <LayersControl.Overlay name='Feature group'>
    <FeatureGroup color='purple'>
      <Popup>
        <span>Popup in FeatureGroup</span>
      </Popup>
      <Circle center={[51.51, -0.06]} radius={200} />
    </FeatureGroup>
  </LayersControl.Overlay>
</LayersControl>
```

Please let me know if you have any feedback and/or chance to try it out, it needs more testing before release but I think the API works like this.
Thanks!


boromisp:
I like the API, but I'm not sure about the implementation.
Instead of using ref, the ControlledLayer subclasses could implement addLayer and removeLayer which would call the addBaseLayer, addOverlay and removeLayer from props.
Then the ControlledLayer sets "this" as the layerContainer for its child MapLayer, and everything works as expected.
Also I believe a "default added" prop for the ControlledLayers could be useful, for having some layers turned off by default.

PaulLeCam:
Thanks for the feedback.

Regarding the use of refs rather than `layerContainer`, it is needed because the `ControlledLayer` subclasses are not layers themselves, they are just container components used for lifecycle, so the contained layers need to be attached to layers created higher in the component hierarchy.

About the "default added" prop I don't think it's possible with the current Leaflet API, the methods to add layers only support the layer and its name as arguments.

boromisp:
The "default added" has no separate API, because with Leaflet you can create a layer, and not add it to the map, only the layers control.

I put together a "proof of concept" for my proposal: #136. It has the same API, I only replaced the ref, and added "defaultAdded" prop.

PR #142, opened 2016-04-16
LayersControl does not pass map props to children

Dahlgren:
Map is not passed to children of LayersControl. This breaks the behaviour defined for custom components.

Could not get `unmock` to work properly and it always passed the test due to ChildComponent not being rendered. My failed attempt at unmock looked like this,

```js
jest.dontMock('../src/LayersControl');
jest.dontMock('../src/Map');
jest.dontMock('../src/MapControl');
jest.dontMock('../src/types/bounds');
jest.dontMock('../src/types/children');
jest.dontMock('../src/types/index');
jest.dontMock('../src/types/latlng');
jest.dontMock('../src/types/controlPosition');
jest.dontMock('../src/types/layerContainer');
```

### DIFF
--- a/__tests__/LayersControl.js
+++ b/__tests__/LayersControl.js
@@ -1,0 +1,40 @@
+import Leaflet from 'leaflet';
+import React from 'react';
+import { render } from 'react-dom';
+
+jest.autoMockOff();
+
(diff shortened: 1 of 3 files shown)


PaulLeCam:
Thanks!
I just published it as v0.11.1.

PR #147, opened 2016-04-26
when dynamically setting the size

RassaLibre:
Hey, I am having a little problem when dynamically setting the size of the map. I think something like this should fix it. Basically when you change the size of the map `.invalidateSize()` should be called (according to Leaflet docs) so Leaflet reacts to the change. Otherwise I end up with half loaded tiles. I suggest something like this. What do you think?

### DIFF
--- a/src/Map.js
+++ b/src/Map.js
@@ -64,6 +64,12 @@ export default class Map extends MapComponent {
     if (bounds && this.shouldUpdateBounds(bounds, prevProps.bounds)) {
       this.leafletElement.fitBounds(bounds, this.props.boundsOptions);
     }
+    if(this.props.style && prevProps.style){
+      if((this.props.style.width && prevProps.style.width && this.props.style.width !== prevProps.style.width)||
+        (this.props.style.height && prevProps.style.height && this.props.style.height !== prevProps.style.height)){
+          this.leafletElement.invalidateSize();
+      }     
+    }
   }
 
   componentWillUnmount() {


PaulLeCam:
Hi,
Thanks for your PR!

I think this behavior is out of the scope of this lib because in your case, your application sets the `height` and `width` styles dynamically, but other style properties could affect, as well as the classes, so effectively completely implementing such a behavior would mean constantly keeping track of the container dimensions.
Considering it is not what Leaflet does itself, instead letting the application call `invalidateSize()` as needed, I think it's better leaving this behavior as part of application logic rather than implementing it in this lib.

RassaLibre:
I see. Well, I am not really sure how can I call `invalidateSize()` on the `Map` object since the object is not exposed to me and I end up with half rendered map. What is your suggestion then?

I guess the main problem is using JSX in this case. I could potentially render the `Map` component to a variable (lets say `renderedMap`), access the function via `renderedMap.leafletElement.invalidateSize()` and then in my JSX use `{renderedMap}`.

PaulLeCam:
The Leaflet element created by each component is accessible using the `getLeafletElement()` method, so you should be able to wrap your `<Map>` with a `ref` property in a parent component handling this logic.
Note that because React renders the div containing the map before creating the leaflet element, it will only be accessible after the `componentDidMount` lifecycle handler.

RassaLibre:
Aaaaaaaah, that solves my problem! Ok then. Thanks! Then I guess there is no point to talk more about this PR and you can close it. Once again, thanks! :)

texas697:
@RassaLibre 
can you post how handle the dynamic resizing of the map. I am having a hard time with this. thanks

RassaLibre:
Hi @texas697, I do something like this. Not sure if this is the best way but it did the magic for me :-)

    class LeafletMap extends React.Component{

	/**
	*
	*/
	constructor(args){
		super(args);
		this.state = {};
	}

	/**
	*
	*/
	componentDidMount(){
		this.refs.map.leafletElement.invalidateSize();
	}

	/**
	*
	*/
	componentDidUpdate(prevProps, prevState){
		if(this.props.width !== prevProps.width)
			this.refs.map.leafletElement.invalidateSize();
	}

	/**
	*
	*/
	render(){
		return(
	        <Map
	        	ref="map"
	        	style={{width: this.props.width, height: this.props.height}}>
	          		<TileLayer url="http://{s}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}.png"/>
	        </Map>
		);
	}
}

Please note that string refs are outdated as they can lead to memory leaks. Use functions instead. Hope it helps!

kjellski:
I just found this and wanted to share a usable version of this from our codebase - this is our drop in replacement wrapper for that, we're using `'react-resize-detector'` for the partable container size detection: 

```import React, { useEffect, useRef, useState } from 'react'
import PropTypes from 'prop-types'
import { Map, TileLayer } from 'react-leaflet'

// NOTE: Needed to enable default marker.
// This is a known issue in react-leaflet: https://github.com/PaulLeCam/react-leaflet/issues/453#issuecomment-410450387
import L from 'leaflet'

import { getHereMapsTileUrl } from '~/const/here-maps'
import { withResizeDetector } from 'react-resize-detector'

delete L.Icon.Default.prototype._getIconUrl
L.Icon.Default.mergeOptions({
  iconRetinaUrl: require('leaflet/dist/images/marker-icon-2x.png'),
  iconUrl: require('leaflet/dist/images/marker-icon.png'),
  shadowUrl: require('leaflet/dist/images/marker-shadow.png'),
})

const defaultMapViewport = {
  center: [53.5513, 9.9869],
  zoom: 13,
}

const defaultMaxZoom = 18
const defaultMinZoom = 3
const defaultStyle = {
  position: 'absolute',
  top: 0,
  left: 0,
  height: '100%',
  width: '100%',
}

const MyMap = ({
  children,
  height,
  maxZoom,
  minZoom,
  style,
  viewport: viewportOverride,
  width,
  ...props
}) => {
  const hereTileUrl = getHereMapsTileUrl()

  const [viewport, setViewport] = useState({
    ...defaultMapViewport,
    ...viewportOverride,
  })

  useEffect(() => {
    setViewport({
      ...defaultMapViewport,
      ...viewportOverride,
    })
  }, [viewportOverride])

  // NOTE: this is using the internal leaflet element to update the size
  const mapRef = useRef()
  useEffect(() => {
    mapRef.current.leafletElement.invalidateSize()
  }, [width, height])

  const mapContainerStyle = { ...style, ...defaultStyle }

  return (
    <Map
      maxZoom={maxZoom || defaultMaxZoom}
      minZoom={minZoom || defaultMinZoom}
      onViewportChanged={setViewport}
      style={mapContainerStyle}
      viewport={viewport}
      ref={mapRef}
      height={height}
      width={width}
      {...props}
    >
      <TileLayer url={hereTileUrl} />
      {children}
    </Map>
  )
}

MyMap.propTypes = {
  children: PropTypes.node,
  height: PropTypes.number,
  maxZoom: PropTypes.number,
  minZoom: PropTypes.number,
  style: PropTypes.object,
  viewport: PropTypes.object,
  width: PropTypes.number,
}

export default withResizeDetector(MyMap)
```

PR #148, opened 2016-04-27
Add dynamic draggable props to Marker

riolowry:
Toggling draggable status is a useful part of leaflet markers, it is nice to have it exposed via props on the Marker component. Usage example included.


PaulLeCam:
Good idea, thanks!

PaulLeCam:
Published in v0.11.3.

PR #153, opened 2016-05-04
Leaflet 1.0 cleanup

mgmeyers:
@PaulLeCam Here are some minor updates to make things more inline with leaflet-1.0. I haven't seen any other discrepancies in the docs, but there are some new features in leaflet-1.0 that might need new components

* `MultiPolygon` and `MultiPolyline` have been removed; `L.polygon` / `L.polyline` now support multis. If desired for backwards compatibility, I could bring back `MultiPolygon` / `MultiPolyline` and just have them export `Polygon` / `Polyline`
* Changed `this.leafletElement.removeFrom(this.props.map)` to `this.leafletElement.remove()` in `MapControl` per leaflet-1.0 specs
* The `radius` argument to `L.circle` has been deprecated and is now an option property
* `CanvasTileLayer` has been removed


PaulLeCam:
Thanks!

PR #159, opened 2016-05-13
Allow zero value for opacity and zIndex

mdebbar:
`BaseTileLayer` currently doesn't accept an opacity of zero. This PR fixes the issue and lets it accept any number.


mdebbar:
Or maybe we want to allow numeric strings e.g. "0.5" ?

PaulLeCam:
Hi,
Thanks for your PR, indeed the initial check for `opacity` and `zIndex` would make it fail for `0`. I removed this check in v0.11.5 so it should work as expected now.

PR #180, opened 2016-06-27
[leaflet 1.0] Change `removeFrom` to `remove` for `MapControl`

nthtran:
In Leaflet 1.0 `Control` doesn't have `removeFrom` anymore, only `remove`.


PaulLeCam:
Thanks!

Issue #183, opened 2016-06-27
Support "flyTo"


PaulLeCam:
Need to implement `flyTo()` and `flyToBounds()` in `Map`, could be set as a switch to enable with a `useFlyTo: boolean` dynamic prop.
http://leafletjs.com/reference-1.0.0.html#map-flyto

PR #192, opened 2016-07-23
Support "flyTo"

jgimbel:
proposed solution to #183

### DIFF
--- a/docs/Components.md
+++ b/docs/Components.md
@@ -81,6 +81,7 @@ This is the top-level component that must be mounted for children ones to be ren
 - `className: string` (optional): className property of the `<div>` container for the map.
 - `maxBounds: bounds` (optional)
 - `style: object` (optional): style property of the `<div>` container for the map.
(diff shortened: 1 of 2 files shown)


PaulLeCam:
Thanks!
Could you please remove the code related to generating an `id` for the map? It should not be needed anymore as the `L.map()` factory accepts a `DOMElement` in Leaflet v1.

jgimbel:
Yep. I tried without and it didn't pass the server side render test. Let me double check that I have leaflet 1 and see if that fixes it after the change.

jgimbel:
Please have a look at the Travic CI output. Should the test be updated? It would seem that v1 does not add its own map id in v1 ([source](https://github.com/Leaflet/Leaflet/blob/v1.0.0-rc.2/src/map/Map.js#L710) and [related source](https://github.com/Leaflet/Leaflet/blob/v1.0.0-rc.2/src/dom/DomUtil.js#L17)).

PaulLeCam:
Oh yes sorry I forgot to push an update to the test, should be OK now.

PaulLeCam:
Thanks!

PR #218, opened 2016-09-08
Updated MultiPolygon 'polygons' prop to allow polygons with holes

ernoaapa:
Like 'Polygon' already allows passing array of arrays to make holes, also
MultiPolygon should allow it. Even though not read in Leaflet
MultiPolygon documentation, it supports it because Polygon supports it.

### DIFF
--- a/src/MultiPolygon.js
+++ b/src/MultiPolygon.js
@@ -8,7 +8,12 @@ import Path from './Path'
 
 export default class MultiPolygon extends Path {
   static propTypes = {
-    polygons: PropTypes.arrayOf(latlngListType).isRequired,
+    polygons: PropTypes.arrayOf(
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Hi,
Thanks for this PR, do you have a reference of this support by Leaflet please?

ernoaapa:
Oh yeah sure! In Leaflet Polygon documentation http://leafletjs.com/reference.html#polygon it says: 
"You can also create a polygon with holes by passing an array of arrays of latlngs, with the first latlngs array representing the exterior ring while the remaining represent the holes inside."

PaulLeCam:
Thanks!

PaulLeCam:
Released in v0.12.2

PR #222, opened 2016-09-18
Plugins.md is updated

yavuzmester:
I have added my recent react-leaflet plugin to the list. The plugin is called react-leaflet-zoom-display and it is to show the current zoom level.


PaulLeCam:
Thanks!

PR #226, opened 2016-09-22
MapPane

rjdestigter:
Developed component implementing Leaflet's new 'panes' concept.

Component name: `MapPane`

Accepted props:
-  `className`
- `style`
- `zIndex`
- `name` (Not required, unique id wiill be created if not passed.)

`Does Leaflet allow to override its existing panes?`
No it does not seem like that. I've allowed passing a "blacklisted" name as prop. However if the name of the pane matches up with a pane leaflet itself creates, it is not removed on `componentWillUnMount`

Subsequently this does allow developers to change the zIndex + style of existing leaflet panes.

Lifecycle:
- `constructor` will create a unique id if no name is passed
- `componentWillMount` will check if the pane exists or not and call `map.createPane` if not.
- `componentDidMount` will set the states `isMounted` to true and update the (created) pane's style
- `componentWillUnmount` will check if the pane's name is blacklisted. If not it will remove the pane from the DOM and leaflet.


rjdestigter:
@PaulLeCam Would you prefer a PR with just the changed and added source code and not a 'built' version?

rjdestigter:
No merge at the moment, pane should should only be passed as an option if it is defined. Leaflet does not override an undefined pane with it's default pane for some of the items (tileLayer for example.)

PaulLeCam:
Thanks for this PR!

I'll review it in more details later, but a few things already:

- Please don't add the transpiled files, it makes the changes more noisy to review.
- Please make sure ESLint checks pass (see Travis failure).
- Could be done in a separate PR, but could you please also add docs + at least one example usage?

Thanks!

PR #227, opened 2016-09-23
Leaflet Map Panes

rjdestigter:
Developed component implementing Leaflet's new 'panes' concept.

Component name: `MapPane`

Accepted props:
-  `className`
- `style`
- `name` (Not required, unique id wiill be created if not passed.)

`Does Leaflet allow to override its existing panes?`
No it does not seem like that. I've allowed passing a "blacklisted" name as prop.

Lifecycle:
- `constructor` will create a unique id if no name is passed
- `componentDidMount` will check if the pane exists or not and call `map.createPane` if not.
- `componentWillUnmount` will remove the pane from the DOM and map

- Added documentation and example.


rjdestigter:
Conflicts with `next` have been resolved.

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+    this.setStyle()
+  }
+
+  componentWillUnmount () {
+    const { name } = this.props
+
+    if (name) {
+      const _name = (name + '').replace(/-*pane/gi, '')
+
+      if (indexOf(BLACKLIST, _name) >= 0) {
```
PaulLeCam: Shouldn't be needed as the the creation of panes with existing names should be prevented.

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+
+    // Remove the created pane
+    const pane = this.getPane()
+    pane && pane.remove &&
+    pane.remove()
+
+    const map = this.context.map || this.props.map
+
+    if (this._name && map && map._panes) {
+      map._panes = omit(map._panes, this._name)
```
PaulLeCam: Seems a bit of a hack, is there any other way to cleanup the pane from Leaflet using more public APIs?

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+import paneType from './types/pane'
+
+const BLACKLIST = ['tile', 'shadow', 'overlay', 'map', 'marker', 'tooltip', 'popup']
+
+export default class MapPane extends Component {
+  static propTypes = {
+    name: PropTypes.string,
+    children: childrenType,
+    map: mapType,
+    zIndex: PropTypes.number,
```
PaulLeCam: `zIndex` should be part of the `style`, unless there is a particular reason to treat it differently?

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+        map.createPane(this._name, this.getParentPane())
+      }
+    }
+  }
+
+  componentWillReceiveProps (props) {
+    this.setStyle(props)
+  }
+
+  setStyle ({ style, zIndex } = this.props) {
```
PaulLeCam: Shouldn't need that, React can render the styles itself.

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+    if (this._name && map && map.createPane) {
+      const existing = this.getPane()
+
+      if (!existing) {
+        map.createPane(this._name, this.getParentPane())
+      }
+    }
+  }
+
+  componentWillReceiveProps (props) {
```
PaulLeCam: Shouldn't be needed.

src/MapComponent.js
```diff
@@ -72,4 +72,17 @@ export default class MapComponent extends Component<any, any, any> {
     const el = this.leafletElement
     if (el) el.fire(type, data)
   }
+
+  getInstanceOptions (props = {}) {
```
PaulLeCam: Could be simply named `getOptions()`

src/MapComponent.js
```diff
@@ -72,4 +72,17 @@ export default class MapComponent extends Component<any, any, any> {
...
+    const pane = props.pane || this.context.pane
```
PaulLeCam: `pane` should always be injected from the context, not props, that could lead to weird behaviors I think.

src/MapPane.js
rjdestigter: There isn't at the moment as far I can find. Browsing Leaflet's source code I could only find:
- `createPane`, 'getPane', and 'getPanes'

Calling L.DomUtil.remove(pane) only removes it from the DOM but not the references leaflet keeps in `map._panes` and `map._paneRenderers`

I had to dig for `_paneRenderers`. It was keeping a reference to the detached pane node causing leaflet to render layers inside it if the pane was added back again.

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+
+    const map = this.context.map || this.props.map
+
+    if (this._name && map && map._panes) {
+      map._panes = omit(map._panes, this._name)
+      map._paneRenderers = omit(map._paneRenderers, this._name)
+    }
+  }
+
+  getParentPane () {
```
rjdestigter: I've added `props.pane` to allow overriding `context.pane`. This is usefull when rendering Popup components that might get an incoming 'pane' as context but you want them to render inside another or leaflet's default poup pane.

Leaflet allows nested panes. `getParentPane` passes a parent pane node to `createPane` if that is the case.

docs/API.md
```diff
@@ -30,6 +30,7 @@ Check Leaflet documentation for the events associated to each component.
   - [BaseTileLayer](Components.md#basetilelayer)
   - [Path](Components.md#path)
 - [Map](Components.md#map)
+- [MapPanes](Components.md#map-pane)
```
rjdestigter: - [ ] Done

docs/Components.md
```diff
@@ -11,6 +11,7 @@ You can directly access the Leaflet element created by a component using `this.l
   - [MapLayer](#maplayer)
   - [Path](#path)
 - [Map](#map)
+- [MapPane](#map-pane)
```
rjdestigter: - [ ] Done

src/types/pane.js
```diff
@@ -0,0 +1,4 @@
+/* @flow */
+import { PropTypes } from 'react'
+
+export default PropTypes.string
```
rjdestigter: - [x] Done

src/index.js
```diff
@@ -24,3 +24,4 @@ export TileLayer from './TileLayer'
 export Tooltip from './Tooltip'
 export WMSTileLayer from './WMSTileLayer'
 export ZoomControl from './ZoomControl'
+export MapPane from './MapPane'
```
rjdestigter: - [x] Done

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+import React, { PropTypes, Component } from 'react'
+import { forEach, omit, uniqueId, indexOf } from 'lodash'
+
+import childrenType from './types/children'
+import mapType from './types/map'
+import paneType from './types/pane'
+
+const BLACKLIST = ['tile', 'shadow', 'overlay', 'map', 'marker', 'tooltip', 'popup']
+
+export default class MapPane extends Component {
```
rjdestigter: - [x] Done

src/MapPane.js
rjdestigter: Syntactic sugar. Since the main purpose of leaflet panes is to have more control over of zIndexes I included it as a prop. Let me know if you want this dropped.

src/MapPane.js
```diff
@@ -0,0 +1,158 @@
...
+  };
+
+  constructor ({name}) {
+    super()
+
+    this.state = {
+      _isMounted: false,
+    }
+
+    this._name = name || `pane-${uniqueId()}`
```
rjdestigter: - [x] Done

src/MapPane.js
rjdestigter: Yes, but React in this case is not in charge of the DOM node right?

src/MapPane.js
rjdestigter: It is prevented by Leaflet. Not preventing in the Pane component allows for using Leaflet's default panes in Pane as well.

src/MapComponent.js
rjdestigter: - [x] Done

src/MapPane.js
rjdestigter: Since React is unaware of the actual DOM node, I have to apply incoming style changes to the pane node.

src/MapPane.js
rjdestigter: This does create a discrepancy if say I have `<Pane name='myCustomPane' />` in two places. If one is unmounted, the other React component might still be alive but it's referencing pane is unmounted. We could either:

- Non-default pane names must be unique and so the pane must not exist yet. `map.getPane(name) should return nothing before `map.createPane` is called for non-default pane names.
- Or we keep track of how often a non-default pane is used somehow.

src/MapPane.js
PaulLeCam: Yes I think it's better to just rely on `style`, I prefer to make the components as "dumb" as possible with no specific logic and let the applications implement custom behavior on top of them.

src/MapPane.js
PaulLeCam: I see, yes I missed the fact Leaflet is creating a new DOM node rather than using the one provided. It's too bad it doesn't allow it as we could have completely leveraged React for this, but in that case I think your implementation is needed as such.

src/MapPane.js
PaulLeCam: Yes, sorry I missed that from Leaflet's `createPane()` implementation.

src/MapPane.js
PaulLeCam: I don't like the idea of allowing to update Leaflet's default panes unless there is a way to restore them to their original styles, that would create unexpected behavior, so I'd rather have this library simply prevent from overriding these panes.
Regarding having "`<Pane name='myCustomPane' />` in two places", same I don't think it should be possible, even if it is technically possible I think it would only lead to unexpected behavior when trying to set different styles in each component for example.

src/MapPane.js
PaulLeCam: OK thanks for the explanation, hopefully Leaflet won't remove these anytime soon!

src/MapPane.js
PaulLeCam: Makes sense, thanks.

PaulLeCam:
@rjdestigter thanks for the changes and explanations.
My only major concern remaining is with allowing to override an existing pane, whether it's a default Leaflet one or another one already created, but I'm happy to merge this PR as-is and work on it myself if you prefer?

rjdestigter:
I suggest we still allow using default leaflet panes but don't set their style in that case. Pane doesn't actually really override existing panes, it just receives them from `map.getPane` and thus won't create them since it already exists.

Todo's:
[-] Remove zIndex prop
[-] Disallow usage of existing panes unless they are default leaflet panes
[-] Don't apply style to default panes. (Disallow style props for default panes)

rjdestigter:
I also realised `className` was being completely ignored so I added that in as well. Proper behaviour was added to deal with `this.props.name` changes.

Also, yes, let's disallow using default names. I thought I needed that feature, but if I want other components such as Popup or any component for that matter to be added to a default pane even if it is nested inside a Pane component, I can just use prop.pane for them.

```
<Pane name='paris'>
  <GeoJSON data={...}>
    <Popup pane={'popupPane'}
```

I will disallow default pane names.

rjdestigter:
Ok, done, using default pane names or that of existing panes will throw an error.

rjdestigter:
`but I'm happy to merge this PR as-is and work on it myself if you prefer?`, I'm good with that if you approve of today's changes. They only thing that I'd like to revisit myself is `getChildren`, but you are welcome to that as well.

I'm not sure if React.cloneElement is really necessary.

rjdestigter:
Also, thanks by the way for letting me collaborate on this project!

rjdestigter:
I acutally went ahead and removed the usage of `cloneElement`. I tested it using nested Panes and it's working fine. 

Also when `map.createPane` is called I store the used name in `this.paneName`. This was necessary because in some cases `componentWillUnmount` is called before `this.setState({name})` is persisted to local state causing `Pane` not to remove the pane.

PaulLeCam:
Thanks for this PR!
I'll need to play a bit with it to make sure I catch possible edge cases before releasing, but hopefully it should make it to the next RC.

PR #229, opened 2016-09-27
Update default image path for 1.0.0

jacobmischka:
Removed rc3 and added trailing /

Also build, looks like that hadn't been done prior to last commit

Without this commit, cdnjs image links were failing as they kept pointing to nonexistent urls like `.../imagesmarker-icon.png` when using leaflet 1.0.0.

I understand if you'd like to fold this change into the rest of the upcoming 1.0.0 changes, but I needed it now and did it already so figured I might as well make a PR.


PaulLeCam:
Hi,
Thanks for this PR but it should no longer be necessary to set the default icon path with Leaflet v1.0.0.

BaptWaels:
Hi,

> it should no longer be necessary to set the default icon path with Leaflet v1.0.0

I use : 
- "leaflet": "^1.0.1"
- "react-leaflet": "^1.0.0-rc.2"

but I still get this error :

.../imagesmarker-icon.png Failed to load resource: 404 (Not Found)
.../imagesmarker-icon.png Failed to load resource: 404 (Not Found)


Why do I still get it ?

jacobmischka:
Currently the path is set with some weird CSS magic in leaflet 1.0. If you're using a module bundler like webpack it probably doesn't support it: https://github.com/webpack/css-loader/issues/322

BaptWaels:
Yes, It tries to load the image from : 

http://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/imagesmarker-icon.png 404 (Not Found)

But the image doesn't exist

jacobmischka:
That line was removed from react-leaflet in 1.0.0-rc.1: https://github.com/PaulLeCam/react-leaflet/commit/b0e95ff0d45b88a5e6d995114642d3114dc4b063#diff-1fdf421c05c1140f6d71444ea2b27638L34

Because the version number you see is still 0.7.7 it looks like you're using a version of react-leaflet older than 1.0.0.rc-2.

BaptWaels:
Oh great ! I was also using react-leaflet-draw. This package has installed react-leaflet 0.12.1 as dependencies.

Everything is alright now.
Thanks a lot for your help :)

PR #232, opened 2016-10-05
Rename GeoJson.js to GeoJSON.js

graf:
Just a small fix in file name; 
It does not build on heroku without it


PaulLeCam:
Thanks!

PR #235, opened 2016-10-11
IE <= 10 support.

### LINKED ISSUE #215: The library is not working with IE10 , v0.12.1 , but v0.11.7 is fine
Looks like the changes in v.0.12.0 moving from props to context broke it on IE10. 
It works fine with v0.11.7

The map is not initialized

The error message is :
 ReactCompositeComponent.getChildContext(): childContextTypes must be defined in order to use getChildContext(). 

You can run the react-leaflet examples locally and see that none are working .

Not sure why the thing is not working. Babel should properly transpile from the ES7 static childContextTypes

dettier:
This PR fixes #215 and returns IE <=10 support.
I turned on loose mode and added transform-proto-to-assign transform.
This is pretty safe to do for this library as it doesn't rely on that specs that are loosened.
As a side-effect bundle size went down 12%.

PS. If you extend ReactLeaflet classes in IE10 you still need to manualy "inherit" static properties:
```
// for IE9 & IE10 support
static contextTypes = BaseClass.contextTypes;
static childContextTypes = BaseClass.childContextTypes;
```


PaulLeCam:
Thanks!

PR #238, opened 2016-10-27
Update leaflet image path

jdeniau:
a `/` was missing at the end of the image path

I also updated the version to 1.0.1 (last version of leaflet)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,4 +33,4 @@ export const setIconDefaultImagePath = path => {
   Leaflet.Icon.Default.imagePath = path
 }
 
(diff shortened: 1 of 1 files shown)


PaulLeCam:
v0.12.x of React-Leaflet only supports Leaflet v0.7.x, please use the v1 RC if you want support for Leaflet v1.

jdeniau:
OK well, I got another bug now, but I [opened an issue](http://stackoverflow.com/questions/40300772/react-leaflet-1-x-create-react-app-404-on-marker-image) on stackoverflow

PR #242, opened 2016-10-28
Update of package to correct path of images

luzeduardo:


luzeduardo:
Will do both commit in the same PR

PR #243, opened 2016-10-28
Update o lint promisse, version number and path for marker images

luzeduardo:


PaulLeCam:
v0.12.x of React-Leaflet only supports Leaflet v0.7.x, please use the v1 RC if you want support for Leaflet v1.

PR #244, opened 2016-10-28
Update of lint promisse and image path correction

luzeduardo:
It's necessary to correct path of image markers


luzeduardo:
I just corrected the image path. Can you merge this?

PaulLeCam:
The path works fine with Leaflet v0.7.x, please check you're using compatible versions of Leaflet and React-Leaflet.
If the issue persist please provide an example using jsfiddle or similar.

PR #250, opened 2016-11-06
Remove flow errors/warnings

joeyrosztoczy:
Hello! We're using react-leaflet for an internal tool where I just added flow and wanted to help update react-leaflet. This is my first  attempt to contribute to an OS library, so please let me know if there's any feedback you have for me.

### Main changes:
- ignore node_modules/fbjs, this library seems to always be throwing flow errors everywhere I use it (including here)
- add annotations to component functions
- add annotations to component returns
- make sure inherited propTypes are specified

### I also had 2 questions that I googled but wasn't sure of:
1. Should I have built the /src prior to the PR? (I did run tests as well)
2. The last round of flow errors were inherited propTypes. Explicitly listing the inherited ones out felt fairly repetitive, but could also be solved by composing the component hierarchy rather than inheriting (i.e. have Path ```return (<MapLayer>) ``` rather than extend it. I'm curious how you'd feel about a PR that refactors towards composability, which would also simplify the explicit typing needs.


PaulLeCam:
Hi,
Thanks for this PR, looks great!

A few things:
- Could you please change the base branch to `next`? It's the branch I'm currently using for v1 RCs so it is more up-to-date than `master`. If it's too much trouble no worries I can merge to `master` first.
- Could you not add the compiled files, just the source ones are fine.
- Not sure this rule is properly unforced by ESLint but could you please remove semicolons where not necessary?

Thanks!

joeyrosztoczy:
@PaulLeCam Thanks for the feedback. 

I'll make all those changes this evening and resubmit! 

I think I was using a misleading local copy of eslint as well, so I'll get those guys fixed.

PR #251, opened 2016-11-07
Reduce flow errors to 0

joeyrosztoczy:
## Hi Paul. I've reopened the PR against next

------------------------------------------------
Hello! We're using react-leaflet for an internal tool where I just added flow and wanted to help update react-leaflet. This is my first  attempt to contribute to an OS library, so please let me know if there's any feedback you have for me.

### Main changes:
- ignore node_modules/fbjs, this library seems to always be throwing flow errors everywhere I use it (including here)
- add annotations to component functions
- add annotations to component returns
- make sure inherited propTypes are specified


PaulLeCam:
Great, thanks!

PR #259, opened 2016-11-23
add super.componentDidUpdate in WMSTileLayer

YannickBochatay:
In the componentDidUpdate method of WMSTileLayer class, there was no call of super.componentDidUpdate.
So when you changed opacity or zIndex, the layer was reloaded.


PaulLeCam:
Thanks!

PR #260, opened 2016-11-28
Fix docs to match GeoJSON case change.

kaylanm:
This caught me for a second, so hopefully it helps someone else.


PaulLeCam:
Thanks!

PR #270, opened 2017-01-09
ImageOverlays can be linked to Popups and Tooltips

rjdestigter:
- [x] Added `childContextTypes` and `getChildContext` for passing down the `popupContainer`
- [x] Removed `render` to implement `MapLayer`'s `render` function


PaulLeCam:
Thanks!

PR #272, opened 2017-01-16
feat(Map): add callback for map state change events

ducreyna:
Hello,

I added a callback props in Map component in order to notify the client when the map state is changing like zoom, center or size of the leaflet element.

Best regards


PaulLeCam:
Hi,
Thanks for your PR but this is out of the scope of this library, this kind of feature should be added to apps or as an external plugin as it's not a core Leaflet feature.

PR #273, opened 2017-01-16
Render Marker children inside div if necessary.

chexxor:
I am passing an array of children into Marker, and I'm getting a runtime React error:

> Error: Marker.render(): A valid React element (or null) must be returned. You may have returned 
undefined, an array or some other invalid object.

I see you recently changed to use this code in a different place - I'd like to do the same thing here, also.

What do you think? Note that I didn't test this change yet - I'm looking for your comment first.

### DIFF
--- a/src/Marker.js
+++ b/src/Marker.js
@@ -54,6 +54,8 @@ export default class Marker extends MapLayer {
   }
 
   render (): null | React.Element<*> {
-    return this.props.children || null
+    return Array.isArray(this.props.children)
+      ? <div>{this.props.children}</div>
+      : (this.props.children || null)
   }
 }


PaulLeCam:
I wonder what is your use case for multiple children in `Marker`?
It's not directly supported by this lib as `Marker` can only optionally have single `Popup` child.

chexxor:
I'm using an elm-like library, called Pux, which wraps React components. In my code, I pass a single child to Marker, and I verify it is sent to React as a single child, but somewhere it is changed to an Array of one element. I spent hours trying to figure out where this bug lies - my React wrapper, React itself, Leaflet - but I can't figure it out.

So, I tried the next-best approach, which is this - changing Marker to accept multiple children.

PaulLeCam:
Well seems to me the actual problem has nothing to do with this lib, it doesn't seem like the right place to fix it.

chexxor:
A Marker is an HTML element. All HTML elements can have multiple children. How is Marker different?
Also, the type of Marker's `children` attribute implies 1 or more nodes, which this patch implements. That furthers my case that this *is* related to this lib.

PaulLeCam:
`<Marker>` is a React element, not an HTML one. The expected behavior is that it can optionally contain a single `<Popup>` element, if you're application is not respecting this behavior, it would not work as expected.

chexxor:
Should we change the type of the `children` attribute to be an instance of Popup, rather than any HTML node? Sounds like this is the only supported value.

PaulLeCam:
Again, `children` has nothing to do with HTML, I think you're confusing concepts here. The `childrenType` is a generic PropType for components that may handle children nodes, it's not a representation of what children are supported and how.
If anything, this seems like a documentation issue to indicate what children are supported, so in the meantime please refer to Leaflet documentation about the relevant elements, as this library aims to provide the same support for children layers and controls.

chexxor:
Ok, thanks for your consideration. I'll react Leaflet docs, but I hope you add some docs to this lib to clarify how Marker's children works.

PR #280, opened 2017-02-15
Update components list in API.mg

h4:
I've just synchronised links between API.md and Components.md (remove obsoleted and add new items).


PaulLeCam:
Thanks!

PR #287, opened 2017-02-26
Expose onPopupOpen and onPopupClose callback events to Popup

sedenardi:
This PR allows `onPopupOpen` and `onPopupClose` callback functions to be passed through to the `Popup` component. These correspond to the popup events from the Leaflet docs (http://leafletjs.com/reference-1.0.0.html#popup).


PaulLeCam:
As [documented by Leaflet](http://leafletjs.com/reference-1.0.0.html#layer-popupopen), these events are exposed by the layers containing popups, not the popups themselves. This also [matches the implementation](https://github.com/Leaflet/Leaflet/blob/master/src/layer/Popup.js#L114) so I wonder what is the purpose of your PR because it seems to me it would trigger these callbacks when any popup gets opened or closed, which would defeat the purpose of adding them on a single popup in the first place.

sedenardi:
Ahhh got it. Thank you for pointing out my mistake.

PR #307, opened 2017-04-02
Update GeoJSON.js

fastlorenzo:
Fixed GeoJson typo (GeoJSON should be GeoJson)

### DIFF
--- a/src/GeoJSON.js
+++ b/src/GeoJSON.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { geoJSON } from 'leaflet'
+import { geoJson } from 'leaflet'
 import { isFunction } from 'lodash'
 import { PropTypes } from 'react'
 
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Thanks for your PR but this is not a typo, `geoJson` has been replaced by `geoJSON` in Leaflet v1 (though the `geoJson` still works as an alias to `geoJSON`).

fastlorenzo:
Hi Paul,

I installed it via npm and received the following error while using geojson: GeoJSON is not a function. After changing the call to GeoJson it worked.
I'll check the version of leaflet when I'm back home and update the thread.

Lorenzo Bernardi

PaulLeCam:
Make sure you are using Leaflet v1+, react-leaflet v1 is not compatible with previous versions.

fastlorenzo:
I checked and the leaflet package (that was installed by another package) was indeed outdated. 
Sorry for the convenience and thanks for that great package!

PR #308, opened 2017-04-04
Only extract leaflet events for prop keys when callback exists

awinograd:
Fixes the following when `myCondition` is `falsey`.

`<MapComponent onClick={myCondition ? myCallback : null} />`

### DIFF
--- a/src/MapComponent.js
+++ b/src/MapComponent.js
@@ -42,7 +42,9 @@ export default class MapComponent extends Component<any, any, any> {
     return reduce(keys(props), (res, prop) => {
       if (EVENTS_RE.test(prop)) {
         const key = prop.replace(EVENTS_RE, (match, p) => p.toLowerCase())
-        res[ key ] = props[ prop ]
+        if (props[ prop ]) {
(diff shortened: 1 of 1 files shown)


src/MapComponent.js
```diff
@@ -42,7 +42,9 @@ export default class MapComponent extends Component<any, any, any> {
     return reduce(keys(props), (res, prop) => {
       if (EVENTS_RE.test(prop)) {
         const key = prop.replace(EVENTS_RE, (match, p) => p.toLowerCase())
-        res[ key ] = props[ prop ]
+        if (props[ prop ]) {
```
awinograd: another option would be to do:

```
if (typeof props[ prop ] === 'function') 
```

awinograd:
The CI seems to be failing for non PR related reasons

PaulLeCam:
Good idea, thanks!

awinograd:
Thanks @paullecam! Hope it's helpful to others as well

PaulLeCam:
Yes, it's always a nice feature to have!
It's now published in v1.1.5.

Issue #319, opened 2017-04-27
Polygon.updateLeafletElement does not properly compare positions to conditionally update leaflet element


resson-ntranquilla:
Please make sure to check the following boxes before submitting an issue. Thanks!

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v15.0.0+).
- [x] Using the supported version of Leaflet (v1.0.0) and its corresponding CSS file is loaded.
- [x] Using the latest version of React-Leaflet.
- [x] The issue has not already been reported.
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/README.md) and that you understand the [technical considerations](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/How%20it%20works.md#limitations).

### Expected behavior
The Polygon element updates when there are new positions set in the "positions" property, as in the `src/Polygon.js` snippet below

```javascript
updateLeafletElement(fromProps: Object, toProps: Object) {
    if (toProps.positions!===fromProps.positions) {
      this.leafletElement.setLatLngs(toProps.positions);
    }
    this.setStyleIfChanged(fromProps, toProps);
  }
```

### Actual behavior

The Polygon element does not update when new positions are set.

### Steps to reproduce

Try to render the Polygon positions dynamically. 

### How to fix
Find a way to properly compare objects. Suggestion below:

```javascript
import { isEqual } from 'lodash' 

[...]

updateLeafletElement(fromProps: Object, toProps: Object) {
    if (!isEqual(toProps.positions,fromProps.positions)) {
      this.leafletElement.setLatLngs(toProps.positions);
    }
    this.setStyleIfChanged(fromProps, toProps);
  }
```

resson-ntranquilla:
I would like to submit a pull request for this, and am unable to push to the repository.

PaulLeCam:
Please provide an example using a jsfiddle or similar, I think the issue you're having is that you're mutating the `positions` array rather than providing a different one.

pans0911:
i think i have the same error,if i change one of the array,the polygon will not change ,if i delete or add one in the array,the poloygon will change.
the way my  is 
                        let newPoints = this.state.points;
			let firstPoints = this.state.points;
[...]

PR #320, opened 2017-04-27
Fix for conditional update of Polygon

resson-ntranquilla:
Relates to issue #319 .

The Polygon updateLeafletElement was not properly comparing objects. This PR introduces lodash's isEqual method to accomplish this conditional update of the element properly.


resson-ntranquilla:
Was able to determine that the bug was in my code.

PR #324, opened 2017-05-18
Update leaflet doc reference version

Chris-Petty:
Corrected to current version of leaflet documentation so react-leaflet users are directed to outdated content.


PaulLeCam:
Good idea, thanks!

PR #326, opened 2017-05-19
Popup: fix DOM leak when unmounting an unclosed popup.

### LINKED ISSUE #325: Unmounting a Popup without closing leaks the children
Please make sure to check the following boxes before submitting an issue. Thanks!

- [X] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [X] Using a supported version of React and ReactDOM (v15.0.0+).
- [X] Using the supported version of Leaflet (v1.0.0) and its corresponding CSS file is loaded.
- [X] Using the latest version of React-Leaflet.
- [X] The issue has not already been reported.
- [X] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [X] Make sure you have fully read the [documentation](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/README.md) and that you understand the [technical considerations](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/How%20it%20works.md#limitations).

### Expected behavior

The child components of a popup should be unmounted by calling `unmountComponentAtNode` if the popup is unmounted.

### Actual behavior

The children stay mounted and the Popup content wrapper is leaked. The reason is evident in the code: `unmountComponentAtNode` is called in `removePopupContent`, which is in turn called in `onPopupClose`. If, however, the popup is unmounted without closing, `unmountComponentAtNode` will never be called. I *think* that this might also be the root cause behind #213 

### Steps to reproduce

I could not provide a working JSFiddle example as I could not get the react devtools working with JSFiddle's iframe. However, using the devtools in the described constellation in a working setup clearly shows that the "detached" children remain registered with React and leak.

DirtyHairy:
Fixes #325 .

### DIFF
--- a/src/Popup.js
+++ b/src/Popup.js
@@ -69,6 +69,8 @@ export default class Popup extends MapComponent {
   }
 
   componentWillUnmount() {
+    this.removePopupContent();
+
     this.context.map.off({
       popupopen: this.onPopupOpen,
       popupclose: this.onPopupClose,


PaulLeCam:
Good catch, thanks!

PR #328, opened 2017-05-21
Allow ImageOverlay bounds to be changed on the fly

Gjum:
Looks like this was just an oversight.

The additional `latLngBounds` cast is necessary because the props can contain a 2-D Array, but `setBounds` expects a `Bounds` object with `_bounds.getNorthWest` etc.


PaulLeCam:
Thanks!

PaulLeCam:
Now published in v1.2.0

PR #330, opened 2017-05-24
Update Getting started.md

Sventenhaaf:
After struggling with parts of the map floating all over the page, I suggest it's cool to add these lines about including css into the app.


PaulLeCam:
Thanks but this is already indicated [above on the same page](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/Getting%20started.md#leaflet-setup), I can't help it if people decide not to read it.

Issue #329, opened 2017-05-24
Suggestion: special-case handling for "load" event


markhepburn:
Please make sure to check the following boxes before submitting an issue. Thanks!

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v15.0.0+).
- [x] Using the supported version of Leaflet (v1.0.0) and its corresponding CSS file is loaded.
- [x] Using the latest version of React-Leaflet.
- [x] The issue has not already been reported.
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/README.md) and that you understand the [technical considerations](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/How%20it%20works.md#limitations).

### Expected behavior

Include an onload handler when constructing a Map, and have it fire.

### Actual behavior

The handler doesn't fire.

### Steps to reproduce

[Example fiddle](https://jsfiddle.net/markhepburn/werszL4d/3/).  This adds onzoomend and onload; open the developer console and notice that onzoomend gets called as expected (when you zoom), but onload is never fired.

This looks to be because the underlying Leaflet map is created before event handlers are bound to it, by which time it has by definition loaded.

Would it be worth adding a special-case in [bindLeafletEvents](https://github.com/PaulLeCam/react-leaflet/blob/d044da9252a51481144167f003224b410d433ea5/src/MapComponent.js#L57) to use [whenReady](https://github.com/Leaflet/Leaflet/blob/7fef7964b85bc2b66867e8287ca1eccbcf383ac2/src/map/Map.js#L1383) instead?

PaulLeCam:
I don't think there should be special cases, it leads to misleading implementations, but I added `whenReady` in v1.2.0.

markhepburn:
Ok, that works for me too, thanks for your prompt response.

PR #331, opened 2017-05-24
Leaflet fires 'load' before handler is bound; use whenReady instead

markhepburn:
See #329 

One possible tweak might be to intercept the handler and add the "type" property to the argument, to be consistent with other handlers, but I haven't done this.


PaulLeCam:
Thanks for your PR but I don't like the idea of changing the expected behavior of Leaflet handlers. Instead, I added a `whenReady` prop to `Map` in v1.2.0 to hopefully cover the use case.

Issue #366, opened 2017-08-23
Polyline 'positions' prop allows invalid prop-types and fails validation on a valid prop-type shape


zemadz:
### Expected behavior

'positions' props validation should succeed for shapes supported by Leaflet and fail for ones that are not supported (http://leafletjs.com/reference-1.2.0.html#polyline). 
```
// should pass validation
const listOfLatLngs = [
  [51.515, -0.09], 
  [51.545, -0.12]
]
const listOfListsOfLatLngs = [
  [
    [51.515, -0.09], 
    [51.545, -0.12]
  ], 
  [
    [51.545, -0.12], 
    [51.335, -0.09]
  ]
]

// should fail validation
const listOfCoordinates = [51.505, -0.09, 51.535, -0.12]
```

### Actual behavior

`listOfCoordinates` passes and `listOfListsOfLatLngs` fails validation.

### Steps to reproduce

No props validation errors should be visible in the logs here (open browser console):
https://www.webpackbin.com/bins/-KsD9ld6Wol9oEUEk5Tw

Props validation error should be visible here:
https://www.webpackbin.com/bins/-KsDBqDmFtZH77T27Pei

PR #367, opened 2017-08-23
Fix Polyline propTypes validation

zemadz:
Fixes https://github.com/PaulLeCam/react-leaflet/issues/366

### DIFF
--- a/src/Polyline.js
+++ b/src/Polyline.js
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types'
 
 import Path from './Path'
 import children from './propTypes/children'
-import latlngList from './propTypes/latlng'
+import latlngList from './propTypes/latlngList'
 import type { LatLng, PathProps } from './types'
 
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Sorry I don't get what you're trying to do... The `latlngList` is already an `arrayOf(latng)`, see https://github.com/PaulLeCam/react-leaflet/blob/master/src/propTypes/latlngList.js#L7 so I don't see in what way your PR is changing anything.

zemadz:
True, `latlngList.js` is `arrayOf(latlng)`, but the import was previously `import latlngList from './propTypes/latlng'` (should have imported `./propTypes/latlng`).

I didn't know/notice that there is a `latlngList.js`, will change my fix to use that.

PaulLeCam:
Ah yes good catch!
Yes if you can please update your PR to fix the import should be good to go.

PaulLeCam:
Should be fixed in v1.6.4

PR #369, opened 2017-08-30
Add react-leaflet-sidebarv2 plugin to the list

markhepburn:
This is a plugin for https://github.com/nickpeihl/leaflet-sidebar-v2, which is a nice collapsable floating sidebar with tabs.  I originally attempted this by wrapping the original plugin, but it does a lot of DOM manipulation and was fundamentally at odds with React, at least how I envisaged it.  So, I render all the markup from scratch and just use the sidebar-v2 CSS, but that seems to work quite well.


PaulLeCam:
Thanks!

PR #388, opened 2017-09-22
Icons from different props are not equal

Fliskr:
Icons from different props are not equal, so we need to check all icons options to set new icon. If you try to use custom icon you make Marker to not be able to be dragged.


PaulLeCam:
Hi,
Thanks for your PR but this comparison is done by reference, it's up to your application to provide the same icon or a different one.

PR #403, opened 2017-10-26
use https for tiles server

sehaas:
Using the Geolocation API in current browsers requires a secure context [0].
I updated the documentation and examples to load the map tiles over secure connections, to prevent possible problems.

[0]https://developer.mozilla.org/en-US/docs/Web/API/Geolocation


PaulLeCam:
Great idea, thanks!

Issue #404, opened 2017-10-30
WMSTileLayer doesn't initialise zIndex or opacity


markhepburn:
**Before** opening an issue, make sure to read the [contributing guide](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/CONTRIBUTING.md) and understand this is a bug tracker, not a support platform.

Please make sure to check the following boxes before submitting an issue.  
**Issues opened without using this template will be closed unless they have a good reason not to follow this template.**

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v15.x or v16.x).
- [x] Using the supported version of Leaflet (v1.2.x) and its corresponding CSS file is loaded.
- [x] Using the [latest stable version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [x] The issue has not already been reported.
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/README.md) and that you understand the [technical considerations](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/How%20it%20works.md#limitations).

### Expected behavior

Should be able to initialise the zIndex and opacity properties on a WMS layer.

### Actual behavior

Those properties are [removed at initialisation time](https://github.com/PaulLeCam/react-leaflet/blob/963a732f0bbd3b1dfb66076606746a92747b77b6/src/WMSTileLayer.js#L24) so not set on the initial element.

(Changes to both are still applied correctly, via updateLeafletElement)

### Steps to reproduce

[WebpackBin](https://www.webpackbin.com/bins/-Kxeju5VZ5VhsVdeqSrD).

markhepburn:
PR incoming.

(First thing I noticed was "404" in the URL and thought "crap, has my ticket just been lost?" :))

PaulLeCam:
Should be fixed in v1.7.2 now thanks to your PR, please try it out.

markhepburn:
Looks great, thanks.

PR #405, opened 2017-10-30
Don't remove zIndex/opacity from initial WMS props

markhepburn:
See PaulLecam/react-leaflet#404.  Initial values were being removed
from the parameters passed to the leaflet element when it was first
created; later updates were still correctly applied.


PaulLeCam:
Thanks!

PR #406, opened 2017-10-30
adding some tests

uxtx:
- viewport options, and change events effecting viewports
- asserting tearDown clears out leaflet element


.gitignore
```diff
@@ -1,6 +1,6 @@
 node_modules
 coverage
-
+.DS_Store
```
PaulLeCam: Please remove

.nvmrc
```diff
@@ -0,0 +1 @@
+6
```
PaulLeCam: Please remove

.gitignore
uxtx: I'm happy to -but this is a standard cruft meta-data file automagically created in mac environments - without it - most contributors on macs may inadvertently slip up and commit .DS_Store files.

uxtx:
@PaulLeCam updated per your suggestions here - thanks for the feedback.

PaulLeCam:
Thanks!

.gitignore
PaulLeCam: Yes I agree these files are annoying, but I think this should be part of developers own git config rather than polluting projects with this, especially since there are different files based on each OS and IDE.

uxtx: works for me!

Issue #422, opened 2017-12-02
Bug: Popup update called before popup content is actually rendered [React 16]


jeffchan:
* [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
* [x] Using a supported version of React and ReactDOM (v15.x or v16.x).
* [x] Using the supported version of Leaflet (v1.2.x) and its corresponding CSS file is loaded.
* [x] Using the [latest stable version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
* [x] The issue has not already been reported.
* [x] Make sure you have followed th [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
* [x] Make sure you have fully read the [documentation](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/README.md) and that you understand the [technical considerations](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/How%20it%20works.md#limitations).

### Expected behavior

Resulting `<div class="leaflet-popup-content">` Leaflet popup should have >51px width.

### Actual behavior

Resulting `<div class="leaflet-popup-content">` Leaflet popup has 51px width (the minimum width).

### Steps to reproduce

Only happens in React 16.

React 16 (not working): https://www.webpackbin.com/bins/-L-OYIxcJoVp94b2EY9s
React 15 (working): https://www.webpackbin.com/bins/-L-OZdE-2LQe33-6oWYe

| Expected | Actual |
|-------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------|
| ![image](https://user-images.githubusercontent.com/97428/33520732-812f2198-d775-11e7-99ae-23f2c3c8c851.png) | ![image](https://user-images.githubusercontent.com/97428/33520729-6db07bee-d775-11e7-8373-943520d96ffb.png) |

### Diagnosis

It appears that with React Fiber, calls to `ReactDOM.render` are committed after the existing lifecycle finishes. You can see this if you analyze the stack traces.

Fix is to run `leafletElement.update` after the `render` is complete (using the callback).

https://github.com/PaulLeCam/react-leaflet/blob/e53f19919cab9cacb6a9a0a800a2b2831205bc97/src/Popup.js#L129-L139

PR #423, opened 2017-12-02
Update Popup element only after render is complete

jeffchan:
Fixes https://github.com/PaulLeCam/react-leaflet/issues/422

----

### Test plan:
Manual :(

Since we mock out `Leaflet.Popup` I was unable to write a test case. Open to suggestions!


PaulLeCam:
Good catch, thanks!

jeffchan:
No problem. Appreciate the quick review, and thanks for your great work.

PaulLeCam:
Thanks! It's now pushed in v1.7.8.

Issue #108, opened 2015-12-15
map rendering issue


dmcd:
Hi Paul,

I'm trying to use leaflet with https://github.com/davezuko/react-redux-starter-kit but the map is not rendering correctly for me. I'm running node 5.2.0 and my leaflet component looks like this:

```javascript
import React from 'react'
import { Map, Marker, Popup, TileLayer } from 'react-leaflet'

class MyMap extends React.Component {
  constructor () {
    super()
    this.state = {
      lat: 51.505,
      lng: -0.09,
      zoom: 13
    }
  }

  render () {
    const position = [this.state.lat, this.state.lng]
    return (
        <Map center={position} zoom={this.state.zoom}>
          <TileLayer
            url='http://{s}.tile.osm.org/{z}/{x}/{y}.png'
            attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
          />
          <Marker position={position}>
            <Popup>
              <span>A pretty CSS3 popup.<br/>Easily customizable.</span>
            </Popup>
          </Marker>
        </Map>
      )
  }
}

export default MyMap
```

<img width="1417" alt="screen shot 2015-12-15 at 15 47 40" src="https://cloud.githubusercontent.com/assets/4025574/11802347/cf7bc71a-a343-11e5-8a3b-4066c2ca5d8e.png">

angusmack:
Your container div needs and absolute height and width and you need to make sure you include leaflet CSS file. Can you confirm you are doing these steps?

I use this lib in my React app and had the  same issue as you before I did this.

dmcd:
Thanks Angus,

My styling wasn't being applied. I've managed to make it work by applying the styling inside index.html like below but I'm not sure why my styling to the react component wasn't being applied (I'm fairly new to this stuff).

```javascript
<!doctype html>
<html lang="en">
<head>
  <title>React Redux Starter Kit</title>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/leaflet.css">
  <style>
    h1, h2, p {
      text-align: center;
    }
    .leaflet-container {
      height: 400px;
      width: 80%;
      margin: 0 auto;
    }
  </style>
</head>
<body>
  <div id="root"></div>
</body>
</html>
```

jordangarside:
[...]

PR #427, opened 2017-12-17
Add CSS import to getting started docs

kontrollanten:
https://github.com/PaulLeCam/react-leaflet/issues/108


PaulLeCam:
Thanks but this code is meant for demo purposes only, it's not intended to be used as-is.

Issue #415, opened 2017-11-20
Tooltip returns error after "removing" it 


FeNoMeNa:
**Before** opening an issue, make sure to read the [contributing guide](https://github.com/PaulLeCam/react-leaflet/blob/master/CONTRIBUTING.md) and understand this is a bug tracker, not a support platform.

Please make sure to check the following boxes before submitting an issue.  
**Issues opened without using this template will be closed unless they have a good reason not to follow this template.**

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v15.x or v16.x).
- [x] Using the supported version of Leaflet (v1.2.x) and its corresponding CSS file is loaded.
- [x] Using the [latest stable version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [x] The issue has not already been reported.
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/README.md) and that you understand the [technical considerations](https://github.com/PaulLeCam/react-leaflet/blob/master/docs/How%20it%20works.md#limitations).

### Expected behavior

The leaflet tooltip should be closed successfully after component unmounting (e.g. when someone change the app route, /map --> /settings).

### Actual behavior

After component unmount I'm getting the following error

```
warning.js:33 Warning: unmountComponentAtNode(): The node you're attempting to unmount was rendered by React and is not a top-level container. Instead, have the parent component update its state and rerender in order to remove this component.

warning.js:33 Warning: unmountComponentAtNode(): The node you're attempting to unmount was rendered by React and is not a top-level container. Instead, have the parent component update its state and rerender in order to remove this component.

leaflet-src.js:6374 Uncaught TypeError: Cannot read property '_targets' of null
    at NewClass.removeInteractiveTarget (leaflet-src.js:6374)
    at NewClass.closeTooltip (leaflet-src.js:10122)
    at NewClass.unbindTooltip (leaflet-src.js:10045)
    at Tooltip.componentWillUnmount (Tooltip.js:95)
    at callComponentWillUnmountWithTimer (react-dom.development.js:8760)
    at HTMLUnknownElement.callCallback (react-dom.development.js:540)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:579)
[...]

PR #443, opened 2018-02-19
🚧 Fix - remove tooltip when interactive

jonathanchrisp:
## Overview
If you create a `<Tooltip>` component which passes the `interactive` and `permanent` props as true, when the component is unmounted it never removes the tooltip from the map. Currently, the code only unbinds the tooltip if `interactive` is not true.

When the parent component's `componentWillUnmount` function is called it simply [removes](https://github.com/PaulLeCam/react-leaflet/blob/master/src/MapComponent.js#L39#L46) the event handlers. 

To give you an example of how I've implemented the functionality and been able to spot the bug please see below: 

```js
return (
  <CircleMarker
    center={center}
    key={key}
    onClick={() => onNodeClick(id)}
    onContextmenu={() => !activeNodeId && setActiveTooltipId(id)}
    pane="markerPane"
    {...options}
  >
    {hasActiveTooltip && (
      <Tooltip
        direction="right"
        interactive
        offset={[20, 0]}
        opacity={1}
        permanent
      >
        <ConnectedNodesList
          addEdge={addEdge}
          removeEdge={removeEdge}
          data={connectedNodeData}
          setActiveEdgeId={setActiveEdgeId}
        />
      </Tooltip>
    )}
  </CircleMarker>
)
```

I spotted that the condition was added to the following [commit](https://github.com/jonathanchrisp/react-leaflet/commit/c625c83a15fecf9a83f63c05a0064cdd288c72d6). If possible can you let me know why this was added? Did this fix a previous bug?

## Example 
tooltip-present

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -70,9 +70,8 @@ export default class Tooltip extends MapComponent<LeafletElement, Props> {
       tooltipopen: this.onTooltipOpen,
       tooltipclose: this.onTooltipClose,
     })
-    if (this.props.interactive !== true) {
-      popupContainer.unbindTooltip(this.leafletElement)
-    }
+
+    popupContainer.unbindTooltip(this.leafletElement)
 
     super.componentWillUnmount()
   }


PaulLeCam:
Hi, the original issue was https://github.com/PaulLeCam/react-leaflet/issues/415
I think it's related to the way components get unmounted, in your case the tooltip is removed but not its container so it doesn't get properly removed, but in the previous issue the container gets removed as well and therefore the unbinding fails to reference it.

It's possible the implementation may have changed in Leaflet 1.3 and fixes the original issue, but that need to be checked before reverting this change.

jonathanchrisp:
Hey @PaulLeCam, I've had another look at this today. I double checked to see if I could replicate the original issue raised in #415 and can't. The issue above appears regardless of which leaflet version you use.

```
leaflet - 1.2.0 & react-leaflet 1.7.4 - no issue
leaflet - 1.3.1 & react-leaflet 1.7.4 - no issue
leaflet - 1.2.0 & react-leaflet 1.8.0 - issue
leaflet - 1.3.1 & react-leaflet 1.8.0 - issue
```

However, just to note we're not using React 16 within our project, we're still using React ^15.6.0 so perhaps this could be why I can't replicate the problem above. For now, we're going to revert to react-leaflet 1.7.4, I'll see if I can create some jsfiddles tomorrow to test the various dependencies, I'll let you know if I find out more.

PaulLeCam:
@jonathanchrisp could you check if v1.8.1 fixes your issue please? I changed the unmounting check, hopefully it covers both cases now.

jonathanchrisp:
@PaulLeCam Tested v1.8.1 and looks good. Thanks for sorting this out 🎉 👍

PR #446, opened 2018-02-20
New react-leaflet plugins

mhasbie:
New react-leaflet plugins:
- react-leaflet-deflate
- react-leaflet-measure
- react-leaflet-easyprint


PaulLeCam:
Thanks!

PR #449, opened 2018-02-22
🚧 Fix - added es folder to files array in package.json

jonathanchrisp:
## Updated
Following the latest version, I've been getting import unresolved errors when running eslint locally. 

```
error  Unable to resolve path to module 'react-leaflet'  import/no-unresolved
```

After checking `node_modules` I noticed that the `es` folder is being cleared out on publish. I've simply added this to the files array.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "url": "https://github.com/PaulLeCam/react-leaflet/issues"
   },
   "homepage": "https://github.com/PaulLeCam/react-leaflet",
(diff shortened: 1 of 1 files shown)


jonathanchrisp:
Hey @PaulLeCam, is there any chance we could merge this in the next few days? If you need me to version I'd more than happy to do so. No worries if not, as I can downgrade back to 1.7.4.

PaulLeCam:
Thanks! Yes should be released soon.

PR #452, opened 2018-02-27
Feature - Support Map handlers when component updated

jonathanchrisp:
## Overview
Currently, if an end user changes any of the top level handler options in the `Map` component the associated Leaflet handlers aren't updated when the component updates. 

In our own use case, we switch on and off the `doubleClickZoom` functionality when switching between various modes in our application. I noticed the other handlers were missing so I've included these.

### DIFF
--- a/src/Map.js
+++ b/src/Map.js
@@ -267,6 +267,7 @@ export default class Map extends MapComponent<LeafletElement, Props> {
 
     if (scrollWheelZoom !== fromProps.scrollWheelZoom) {
       if (scrollWheelZoom === true || typeof scrollWheelZoom === 'string') {
+        this.leafletElement.options.scrollWheelZoom = scrollWheelZoom
         this.leafletElement.scrollWheelZoom.enable()
       } else {
         this.leafletElement.scrollWheelZoom.disable()
@@ -283,6 +284,7 @@ export default class Map extends MapComponent<LeafletElement, Props> {
 
     if (touchZoom !== fromProps.touchZoom) {
       if (touchZoom === true || typeof touchZoom === 'string') {
+        this.leafletElement.options.touchZoom = touchZoom
         this.leafletElement.touchZoom.enable()
       } else {
         this.leafletElement.touchZoom.disable()
--- a/src/Map.js
+++ b/src/Map.js
@@ -266,7 +266,7 @@ export default class Map extends MapComponent<LeafletElement, Props> {
     }
 
     if (scrollWheelZoom !== fromProps.scrollWheelZoom) {
-      if (scrollWheelZoom === true || scrollWheelZoom === 'center') {
+      if (scrollWheelZoom === true || typeof scrollWheelZoom === 'string') {
         this.leafletElement.scrollWheelZoom.enable()
       } else {
         this.leafletElement.scrollWheelZoom.disable()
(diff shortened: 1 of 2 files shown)


jonathanchrisp:
Node 6 build failing on Travis CI with: 

```
error @shellscape/koa-static@4.0.4: The engine "node" is incompatible with this module. Expected version ">= 7.6.0".
error Found incompatible module
```

src/Map.js
```diff
@@ -226,6 +233,62 @@ export default class Map extends MapComponent<LeafletElement, Props> {
...
+    if (keyboard !== fromProps.keyboard) {
+      if (keyboard === true) {
+        this.leafletElement.keyboard.enable()
+      } else {
+        this.leafletElement.keyboard.disable()
+      }
+    }
+
+    if (scrollWheelZoom !== fromProps.scrollWheelZoom) {
+      if (scrollWheelZoom === true || typeof scrollWheelZoom === 'string') {
```
PaulLeCam: Why checking if `scrollWheelZoom` is a string?

src/Map.js
```diff
@@ -226,6 +233,62 @@ export default class Map extends MapComponent<LeafletElement, Props> {
...
+    if (tap !== fromProps.tap) {
+      if (tap === true) {
+        this.leafletElement.tap.enable()
+      } else {
+        this.leafletElement.tap.disable()
+      }
+    }
+
+    if (touchZoom !== fromProps.touchZoom) {
+      if (touchZoom === true || typeof touchZoom === 'string') {
```
PaulLeCam: Why checking if `touchZoom` is a string?

src/Map.js
jonathanchrisp: `scrollWheelZoom` accepts a boolean value or a string in the latest leaflet [documentation](http://leafletjs.com/reference-1.3.0.html#map-scrollwheelzoom)

src/Map.js
jonathanchrisp: As above - [link](http://leafletjs.com/reference-1.3.0.html#map-touchzoom)

jonathanchrisp:
@PaulLeCam Updated as discussed. Just to note, I took another look at the Leaflet source code regarding the string value that can be passed for `scrollWheelZoom ` and `touchZoom` and it appears to only check for the 'center' value so I've made the above conditionals stricter.

PaulLeCam:
Thanks!
I think it might needs additional logic in the `scrollWheelZoom` and `touchZoom`, the behavior when setting the value to `center` would enable the handler, but as if the value provided was `true` rather than the expected `center` behavior, unless I'm missing something?

One way to achieve this could be to update the `map.options` Object in the Leaflet element to reflect the change, I don't think it needs to check for the `center` value specifically, checking for the string type is likely a better way to make sure the behavior will also work for other values that could be added by Leaflet later.

jonathanchrisp:
Hey @PaulLeCam, I've reset the conditional logic for `scrollWheelZoom` and `touchZoom` in the latest commit. 

In terms of your first point, I'm not entirely sure what you mean. However, the logic checks whether the changed value is true OR a string type. In terms of the leaflet code, regardless of the value passed (`true` or `center`), we just need to enable/disable the handler. 

It checks the value passed down when performing an action  e.g. scrollWheelZoom's [_performZoom](https://github.com/Leaflet/Leaflet/blob/703ae02aa8cbd0b87be5b01e77754b83ad732267/src/map/handler/Map.ScrollWheelZoom.js#L80#L84)

jonathanchrisp:
Hey @PaulLeCam, apologies, now I know what you mean. Thanks for the detailed explanation above. I've updated the functionality in the latest commit.

jonathanchrisp:
Hey @PaulLeCam, just checking (no rush) but is there anything else to add here?

PaulLeCam:
Looks great, thanks!
Sorry I don't have much time to play with the branch at the moment but I'll try to get to it in the coming days.

PR #454, opened 2018-03-02
Added scrollWheelZoom to updateLeafletElement

felixflow:


jonathanchrisp:
Hey @felixflow, I opened a PR a few days ago which cover this functionality and all the other map handlers, you can see this in #452.

PR #459, opened 2018-03-12
Remove event listeners

andersonveiga:
This prevents an infinite loop error that I experienced on my React app, apparently the events were being fired even after the DOM element was removed


PaulLeCam:
Thanks!

andersonveiga:
No problem 🤗

PR #461, opened 2018-03-14
Added check if popup is open in react render callback function

olabalboa:
I am using react-leaflet together with my own react-implementation of leaflet markercluster. The following error can occurr (see stack trace below).

1. I have a popup open for a marker
2. The content of the popup is changed at the same time as the position of the marker is changed.
3. The position change results in that the marker gets clustered with another marker and the popup gets closed by leaflet-markercluster.
4. Since the content of the popup is changed, the callback function supplied to the render call in renderPopupContent is executed. 
5. this.leafletElement._adjustPan() is called but since the popup no longer exists on the map an error is thrown in the leaflet marker code.

My suggestion is to check if the popup is open in the callback function.
`Uncaught TypeError: Cannot read property '_panAnim' of null
    at NewClass._adjustPan (webpack://%5Bname%5D/./node_modules/leaflet/dist/leaflet-src.js?:9476)
    at eval (webpack://%5Bname%5D/./node_modules/react-leaflet/lib/Popup.js?:83)
    at commitCallbacks (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:6163)
    at commitLifeCycles (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:8793)
    at commitAllLifeCycles (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:9946)
    at HTMLUnknownElement.callCallback (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:542)
    at Object.invokeGuardedCallbackDev (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:581)
    at invokeGuardedCallback (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:438)
    at commitRoot (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:10050)
    at performWorkOnRoot (webpack://%5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js?:11017)`


olabalboa:
Creating a new pull request based on v1.9.0

PR #462, opened 2018-03-20
Adding an "animated" behavior to the marker

alcmoraes:
With this little change, the user will be able to set an "animated" boolean property to the marker that will "simulate" an animated behaviour from the previous position to the next position.


PaulLeCam:
Thanks for your PR but this should be in a separate project, this library only implements core Leaflet functionalities.

PR #466, opened 2018-04-06
assign a ref to the marker (React 16.3)

P45QU10U:


PaulLeCam:
Thanks for your PR but the master branch should be compatible with previous versions of React, could you please change the target branch to `v2`?

P45QU10U:
OK. It's fixed.

PaulLeCam:
Thanks!

PR #475, opened 2018-05-19
Use createPortal for rendering DivOverlays

hannesj:
This makes it possible to natively pass context trough to the subtree that is rendered inside the popup.


PaulLeCam:
Great idea, thanks!

PR #479, opened 2018-06-01
Add react-leaflet-nmscale to plugins list

anajavi:
react-leaflet-nmscale is nautical mile scale component.

works only with ^2.0.0-beta.1 || ^2.0.0


PaulLeCam:
Thanks!

PR #486, opened 2018-06-12
Hotfix/flow errors

cybear:
Hi!

Thanks for a nice code library!
I am proposing to avoid warnings altogether through writing a little more verbose code.

**What's the purpose?**

For projects like the one I'm currently working in, which already has its own `.flowconfig` set the way we like it, this library causes warnings to occur. So your library will be able to reach a larger user base, and cause less annoyance, if the library can adhere to these flow lint checks.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -7,10 +7,6 @@
 [include]
 
 [lints]
-all=warn
-deprecated-type=off
-unclear-type=off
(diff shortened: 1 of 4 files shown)


PaulLeCam:
Closing due to lack of feedback, please reopen if you want to update the changes.

PR #508, opened 2018-08-22
Add react-leaflet-freedraw to plugins list

elangobharathi:
[react-leaflet-freedraw](https://github.com/elangobharathi/react-leaflet-freedraw) is a plugin that integrates [Leaflet.FreeDraw](https://github.com/Wildhoney/Leaflet.FreeDraw) library.


PaulLeCam:
Thanks!

PR #509, opened 2018-08-26
Modified "Lifecycle process" overview in intro.md

clintharris:
1. Updated lifecycle overview to explain that child components create their Leaflet element in the `constructor()` (not `componentWillUpdate()`, which is no longer used).
2. Added an additional step explaining that the component adds its Leaflet element to the map in `componentDidMount()`.


PaulLeCam:
Thanks!

PR #512, opened 2018-08-31
Fixed attributions not being updated when props change.

FizzyGalacticus:
When changing a TileLayer's attribution prop, it was not updating the DOM. This implements a fix using the Leaflet API as documented [here](https://leafletjs.com/reference-1.3.4.html#control-attribution-addattribution).


PaulLeCam:
Hi,
Thanks for your PR but I don't think this behavior should be limited to `TileLayer`, as other layers could also support it.
I added support for dynamic `attribution` prop in v2.1.0 so I hope this covers your use case.

PR #515, opened 2018-09-07
Inform about missing context element

rubenspgcavalcante:
When creating a custom component if it use some class that directly/indirectly extends MapLayer, it will fail if not using withLeaflet HOC (missing leaflet inside its context).
The error doesn't point out to it, I'm throwing a custom error whenever it's missing the leaflet on context, informing about the use of the HOC.

Also, to added the info about the support of my plugin for the v2 👍

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -10,7 +10,7 @@ open pull requests to update this list!
 
 | Plugin                                                                                                           | Compatible with v2 |
 | ---------------------------------------------------------------------------------------------------------------- | ------------------ |
-| [`react-leaflet-ant-path`](https://www.npmjs.com/package/react-leaflet-ant-path)                                 | unknown            |
+| [`react-leaflet-ant-path`](https://www.npmjs.com/package/react-leaflet-ant-path)                                 | **yes**            |
 | [`react-leaflet-bing`](https://www.npmjs.com/package/react-leaflet-bing)                                         | unknown            |
 | [`react-leaflet-choropleth`](https://www.npmjs.com/package/react-leaflet-choropleth)                             | unknown            |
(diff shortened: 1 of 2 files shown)


PaulLeCam:
Thanks for your PR but please remove this runtime check, this is something that should be covered by Flow.

rubenspgcavalcante:
Hi @PaulLeCam,
Sorry but I don't share your point of view. You see, this part can't be covered by flow if you're not using it (and you can't enforce any app/lib, that uses react-leaflet as dependency, to use flow).
So still thinking that the run-time check is valid :)
What do you say? 🤔

rubenspgcavalcante:
Another thing, even if your app proper use flow (I'm not a expert on it tho), don't know if flow will be able to check on the minified/distribution version of react-leaflet

rubenspgcavalcante:
Sorry, but isn't about Flow or not flow, but being more explicit than implicit.
This is the following error when not using `withLeaflet`:
```
Uncaught TypeError: Cannot read property 'layerContainer' of undefined
    at MyComponent.get (MapLayer.js:73)
    at MyComponent.componentDidMount (MapLayer.js:40)
    at commitLifeCycles (react-dom.development.js:14524)
    at commitAllLifeCycles (react-dom.development.js:15738)
    at HTMLUnknownElement.callCallback (react-dom.development.js:145)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:195)
    at invokeGuardedCallback (react-dom.development.js:248)
    at commitRoot (react-dom.development.js:15908)
    at completeRoot (react-dom.development.js:17296)
    at performWorkOnRoot (react-dom.development.js:17224)
```
Without the checking, there's no reason for this PR.
At least it will be **documented** for lost developers, who are extending something from your lib (and I'm not sure if even Flow can check this), that the error above is about the missing HOC.

Just my 2 cents 👍

PR #524, opened 2018-10-04
Add react-leaflet-wmts to plugins page

mpivaa:
I've built a plugin that wraps https://github.com/mylen/leaflet.TileLayer.WMTS behaviour. I think it could be useful to have in the plugins page.

Thanks for react-leaflet by the way!


PaulLeCam:
Thanks!

PR #527, opened 2018-10-16
Use ESM babel runtime functions for jsnext build

anajavi:
Build with ESM versions of babel runtime functions. Results in slightly smaller app bundle size.

docs:
https://babeljs.io/docs/en/babel-plugin-transform-runtime#useesmodules


PaulLeCam:
Thanks!

PR #533, opened 2018-10-28
Replace some lodash functions with ES

radeno:
Replace lodash functions `reduce`, `assign` and `isFunction` with ES alternative.
Not all because `omit` and `pick` hasn't pretty alternative and `isEqual` is too 

Size changes:
Before
```
react-leaflet.js: 194 510 bytes
react-leaflet.min.js: 53 515 bytes
```
After
```
react-leaflet.js: 185 075 bytes
react-leaflet.min.js: 52 676 bytes
```

One question about GeoJSON style checking. `isFunction` was used because `typeof` was different in some older browsers? https://github.com/PaulLeCam/react-leaflet/blob/0f1196b27799eebf14f1c76bfaa4440d51506300/src/GeoJSON.js#L29
If so that browsers are unsupported.
http://www-lia.deis.unibo.it/materiale/JS/developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof.html


src/WMSTileLayer.js
```diff
@@ -32,9 +32,8 @@ class WMSTileLayer extends GridLayer<LeafletElement, Props> {
   }
 
   getOptions(params: Object): Object {
-    return reduce(
-      super.getOptions(params),
-      (options, value, key) => {
+    return Object.entries(super.getOptions(params)).reduce(
```
PaulLeCam: Could you use `Object.keys()` here please? I think there are more browsers supporting it than `Object.entries()`.

radeno:
@PaulLeCam changed

PaulLeCam:
Thanks!

PR #537, opened 2018-11-06
Fix comparison that causes extra renders

pferreir:
This looks to me like an accidental mistake that causes a high number of additional renders. The comparison would always return false, since the object would most likely not be the exact same.

### DIFF
--- a/src/Marker.js
+++ b/src/Marker.js
@@ -1,6 +1,7 @@
 // @flow
 
 import { type Icon, Marker as LeafletMarker } from 'leaflet'
+import { isEqual } from 'lodash'
 import React from 'react'
 
 import { LeafletProvider, withLeaflet } from './context'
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Hi,
Thanks for you PR but this is the expected behavior, it's up to the application logic to handle this, for example leveraging [React's `shouldComponentUpdate()`](https://reactjs.org/docs/react-component.html#shouldcomponentupdate).

pferreir:
Hi,
Thanks for having had a look at the PR. I'm not sure I understand, could you maybe elaborate a bit on how you would obtain the same effect with `shouldComponentUpdate`?
The only workaround that I can think of and doesn't imply patching the `Marker` class is subclassing it.

pferreir:
Hi @PaulLeCam, just pinging you to make sure that you got the last comment. This really looks to me like a bug (comparing `LatLng` objects with `!==`).

PaulLeCam:
Hey sorry I missed this, basically this is how shallow comparison works, similar to React.PureComponent's behavior I think.
The logic doing more custom comparison should likely go one level higher in your own component that renders the Marker.

pferreir:
Thanks a lot for your answer, @PaulLeCam. I see what you mean, I am indeed re-creating the Marker object too often. I still think that avoiding this pitfall would be useful, but I understand your reluctance in going that way. Thanks once again!

PR #545, opened 2018-12-14
Added example for custom icons

rlahoda:
-Created a customIcons component in custom-icons.js
-Added new component to app.js
-Created an example/assets folder
-Added two svg markers that I made in the folder for use as the markers in the example component
-Added mention of custom-icons.js to docs/examples.md
-The marker-shadow is just the same one that comes default with Leaflet but I copied it in to the example/assets folder for simplicity

-the SVG files and PNG files were not loading correctly in Webpack. To fix this I found the webpack file-loader package and added the appropriate test code into the webpack.config file

I hope the Webpack change is ok. I wasn't thrilled to have to add that and I understand if you don't want to change the config and add the extra package, but it was the only way I could figure out to get the images to load. If there's a better/easier way, I'll be glad to make the change.

I really appreciate all the work that's been put into this package and I am glad to be able to add something to it.


rlahoda:
I pulled the icon declarations into another file. This is how I approached it in my own project but was trying to have it in the component for simplicity in the example. Let me know if this works for you. Thanks!

rlahoda:
I created a second version that has the icons declared in the component in their own exports. This might be more straightforward since they show the icons in the same file

rlahoda:
the custom-icons2 isn't referenced anywhere. If you want to keep both, I can create another pr to add mention of the file to the docs md file.

PaulLeCam:
Maybe the example with the icons in the same module could be the single one, what do you think?
As long as the icons are not created in the `render()` method I think it's good either way.

rlahoda:
I'm fine with either. From an example standpoint, having everything in one file makes sense because the user doesn't have to search around for the second file with the icons in it. I didn't think of how to do it all in one file like that until after making the 2-file version. I'd rather the single file version if it's all the same to you. If that's good with you, I'll make the change and resubmit.

rlahoda:
I removed the second custom-icons2.js file and the extra icons.js file and added the two icons into the single custom-icons.js component. Let me know if you'd like any additional changes.

Thanks!

PR #550, opened 2019-01-03
(#549) Remove duplicate element creation

### LINKED ISSUE #549: Duplicate creation of leafletElement in Path
`MapLayer` constructor creates `leafletElement` in its only statement:
```javascript
export default class MapLayer<
...
  constructor(props: Props) {
    super(props)
    this.leafletElement = this.createLeafletElement(props)
}
```
https://github.com/PaulLeCam/react-leaflet/blob/de369a5d8fc9610ebfc909fc036e53d089e2186a/src/MapLayer.js#L18
`Path` component extends `MapLayer`, calls `MapLayer` constructor and then immediately calls the same `createLeafletElement` function and saves the result to the same field `leafletElement`:
```javascript
export default class Path<...> extends MapLayer<LeafletElement, Props> {
...
  constructor(props: Props) {
    super(props)
    this.leafletElement = this.createLeafletElement(props)
    if (this.contextValue == null) {
      this.contextValue = {
        ...props.leaflet,
        popupContainer: this.leafletElement,
      }
    }
  }
```
https://github.com/PaulLeCam/react-leaflet/blob/de369a5d8fc9610ebfc909fc036e53d089e2186a/src/Path.js#L40
Actually, the same data is creating twice. There is no need to create `leafletElement` in `Path`.

vadzim:
closes #549

### DIFF
--- a/src/MapLayer.js
+++ b/src/MapLayer.js
@@ -12,6 +12,7 @@ export default class MapLayer<
   Props: MapLayerProps,
 > extends MapComponent<LeafletElement, Props> {
   contextValue: ?LeafletContext
+  leafletElement: LeafletElement
 
   constructor(props: Props) {
     super(props)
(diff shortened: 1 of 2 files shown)


PaulLeCam:
Thanks!

Issue #553, opened 2019-01-10
Remove Lodash


jtfell:
Thanks for this lib, the external API is an absolute pleasure!

From a quick search of this project I can see that is using the following functions from lodash:

- isEqual
- pick
- forEach
- uniqueId
- omit

Replacing these functions with vanilla JS alternative implementations will both reduce the bundle size (leaflet.js doesn't use lodash at all) and simplify the build (no need for lodash specific plugins).

Would you be open to merging a PR making this change? If so, I'll get started on replacing them. What is your testing procedure like at the moment?

PR #554, opened 2019-01-11
replace lodash with fast-deep-equal and vanilla JS implementations

jtfell:
I've gone ahead and done a draft implementation for #553

### Size comparison (bytes)

**Before**
dist/react-leaflet.js 187021 
dist/react-leaflet.min.js 52367

**After**
dist/react-leaflet.js 82785 (44% of original size)
dist/react-leaflet.min.js 33663 (64% of original size)


PaulLeCam:
Thanks!

PR #564, opened 2019-02-05
Update plugins

mhasbie:
Update plugin compatibility status:
* react-leaflet-deflate
* react-leaflet-easyprint
* react-leaflet-vectorgrid

Add new plugins:
* react-leaflet-dialog
* react-leaflet-magnifying-glass


PaulLeCam:
Thanks!

PR #567, opened 2019-02-12
Update react-leaflet-control compatibility status

robinmetral:
`react-leaflet-control` was updated 5 months ago and is compatible with react-leaflet v2.


ZhangYiJiang:
@robinmetral There is a bug that affects v2 which I'm trying to fix right now https://github.com/LiveBy/react-leaflet-control/issues/27

robinmetral:
Thanks @ZhangYiJiang :rocket: Of course it's still compatible, and there are workarounds in the issue you mentioned. Looking forward to a cleaner fix.

PaulLeCam:
Thanks!

PR #568, opened 2019-02-13
Updates third party compatibility docs

renanbandeira:
I was facing some issues while trying to use react-leaflet-div-icon and react-leaflet-heatmap-layer with react-leaflet, and then I realized that react-leaflet-div-icon [doesn't support react-leaflet v2.0](https://github.com/jgimbel/react-leaflet-div-icon/issues/20) and [react-leaflet-heatmap-layer does](https://github.com/OpenGov/react-leaflet-heatmap-layer/blob/master/CHANGELOG.md#200-release).


PaulLeCam:
Thanks!

PR #570, opened 2019-02-16
Passing LayerGroup options to the correct leaflet LayerGroup parameter

failociraptor:
While working with react-leaflet LayerGroup, I noticed that the annotations were not being passed correctly to the underlying Leaflet LayerGroup.  

None of the options were being passed through correctly, as they're being passed in as the layers instead of the options.


PaulLeCam:
Good catch, thanks!

PR #577, opened 2019-03-11
Draggable Marker example property name fix up.

Dan-Wood:
Hi all

I've fixed a naming mistake on event name property in the Draggable Marker example.

onDragend doesn't exist on MarkerEvents it's ondragend. :)

Cheers
Dan

### DIFF
--- a/example/components/draggable-marker.js
+++ b/example/components/draggable-marker.js
@@ -53,7 +53,7 @@ export default class DraggableExample extends Component<{}, State> {
         />
         <Marker
           draggable={this.state.draggable}
-          onDragend={this.updatePosition}
+          ondragend={this.updatePosition}
           position={markerPosition}
           ref={this.refmarker}>
           <Popup minWidth={90}>


PaulLeCam:
Hi,
Thanks but the event names are normalized in the binding logic so case doesn't matter.

Dan-Wood:
Sorry, but this does kind of matter, especially when using Typescript as the build will fail to compile.

Unless I'm doing this wrong, please correct me here.

`
          <Marker
            draggable={true}
            position={this.state.marker}
            ref={this.refmarker}
            onDragend={this.draggedMarker}
            icon={getMapIcon()}
          />
`

> TS2322: Type '{ draggable: true; position: Position; ref: RefObject<Marker<MarkerProps, Marker<any>>>; onDragend: () => void; icon: any; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Marker<MarkerProps, Marker<any>>> & Readonly<{ children?: ReactNode; }> & Readonly<MarkerProps>'.
>   Property 'onDragend' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Marker<MarkerProps, Marker<any>>> & Readonly<{ children?: ReactNode; }> & Readonly<MarkerProps>'.

And obviously

`
          <Marker
            draggable={true}
            position={this.state.marker}
            ref={this.refmarker}
            ondragend={this.draggedMarker}
            icon={getMapIcon()}
          />
`

Works perfectly fine.

PaulLeCam:
These examples are not using TypeScript, so how is this related?
You can use a fully lowercase `ondragend` in your code if it matters to you or your setup.

Dan-Wood:
Regardless,it's the only you showcase or document how to do a draggable marker with react-leaflet.

And leaflet clearly says use dragend.

Surely you can see the confusion there. I don't see why it matters of your examples use typescript or not.

There's no consistency.

PR #582, opened 2019-04-15
TileLayer customizable create tile

GiovanniMontenegro:
Hi,
I added the possibility to customize the retrieve method for each tile.

This feature is important, to use a custom server to get tile images, behind a proxy.

Regards,
Giovanni

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -139,54 +139,54 @@ React-Leaflet provides the `onViewportChange` and `onViewportChanged` callbacks 
 
 ```javascript
 type Viewport = {
-	center: [number, number],
-	zoom: number,
+    center: [number, number],
+    zoom: number
 };
 
 type Props = {
-	viewport: Viewport,
+    viewport: Viewport
 };
 
 type State = {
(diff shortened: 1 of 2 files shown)


PaulLeCam:
Hi,
Thanks but I think this should be part of a custom component rather than built-in as this is using an internal method from Leaflet rather than a public option.

GiovanniMontenegro:
Yes is an interla method of leaflet, that is public.
If you have a proxy behind, you can't use the built-in method but you have to change it. 
With the implementation of react-leaflet, this possibility is not available.
The component can't be used behind a proxy.

Thanks for your answer,
I will fork your project then.

Regards,
Giovanni

PaulLeCam:
Hi,

You can read about [creating custom components in the docs](https://react-leaflet.js.org/docs/en/custom-components.html).
In your case, extending the `GridLayer` component should be a good start, likely to implement something similar as the [`TileLayer component`](https://github.com/PaulLeCam/react-leaflet/blob/master/src/TileLayer.js).

PR #607, opened 2019-08-05
use deep comparison for Marker position

njunxuan:
Comparing the position array by reference leads to unnecessary updating of the Marker. This leads to very slow performance when there are many markers. I believe this partially fixes https://github.com/YUzhva/react-leaflet-markercluster/issues/78.

Is there some reason this is not the default method of comparison?


PaulLeCam:
Thanks for your PR but it's up to you application or plugin to handle this logic as needed.

PR #617, opened 2019-09-09
Remove unmaintained rotatedmarker plugin

verdie-g:


kjellski:
So sad to see you go... I would love to take over maintenance, but the current api makes it really hard to maintain the plugins I feel.

PaulLeCam:
Thanks!

PR #618, opened 2019-09-11
Avoid useless and issue-prone re-renderings because of shallow props objects comparisons

vala:
This PR uses the already present `fast-is-equal` package and replaces all object comparisons (`obj1 === obj2` or `obj1 !== obj2`) with `isEqual` calls to actually compare their props and not their internal references.

This is needed to avoid leaflet elements re-rendering when props didn't actually change, which may cause third party code or libraries to receive update events and behave unexpectedly.

My case was [react-leaflet-markercluster](https://www.npmjs.com/package/react-leaflet-markercluster) that received a `"move"` event when I clicked a marker as `Marker.prototype.updateLeafletElement` did behave like the location props did change, although there were no difference between the `prevProps` and current `props`. 

I saw some issues / PRs that talked about the issue but people seemed to find other ways to handle the case, but this seem like a good thing to handle properly to avoid unexpected issues, since it took me several hours to track this down.

### DIFF
--- a/src/Circle.js
+++ b/src/Circle.js
@@ -1,6 +1,7 @@
 // @flow
 
 import { Circle as LeafletCircle } from 'leaflet'
+import isEqual from 'fast-deep-equal'
 
 import { withLeaflet } from './context'
 import Path from './Path'
@@ -21,7 +22,7 @@ class Circle extends Path<LeafletElement, Props> {
   }
 
   updateLeafletElement(fromProps: Props, toProps: Props) {
-    if (toProps.center !== fromProps.center) {
+    if (!isEqual(toProps.center, fromProps.center)) {
       this.leafletElement.setLatLng(toProps.center)
     }
     if (toProps.radius !== fromProps.radius) {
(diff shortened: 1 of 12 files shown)


PaulLeCam:
Thanks but as previously discussed in various issues, this is the expected behavior. You can perform shallow comparison at your application or plugin level.

vala:
Hi and thanks for your answer.

I took a look at existing closed issues/PRs talking about this same issue, and didn't find any _discussion_ about the reasons why this is an expected behavior ?

Does this behavior have any positive outcome ? Like allowing to force the re-render of a react-leaflet component by passing a clone of the previous prop to the component ?

Thanks for your time, I hope to be able to better understand the logic behind your current position.

PaulLeCam:
You can read about it in the [documentation](https://react-leaflet.js.org/docs/en/intro#component-properties), stating the properties are compared by reference.

Basically this library only aims to be a bridge between React and Leaflet, with minimum logic in-between to make things work, so if a component is given different props the expected behavior is that it calls the relevant Leaflet methods.
The way Leaflet handles these calls is up to Leaflet, and the way these props are provided to React Leaflet components is up to React and your app.

PR #623, opened 2019-09-30
Simplest example does not work with latlng array

fullhdpixel:
You need to make it into an object Map center and Marker position


PaulLeCam:
I works as expected when running locally and in the [example codepen](https://codepen.io/PaulLeCam/pen/gzVmGw) so it's possible whatever issue you're having comes from your environment.
Please reopen with a PR passing CI and a failing test case if the problem persists.

Issue #624, opened 2019-10-01
SVGOverlay component


fenech:
Hi, I notice that this library doesn't contain a component for Leaflet's SVGOverlay. I made one for my own project - it's written in TypeScript but it would be fairly easy to port to JS/Flow and open a PR if you're interested.

To begin with, I tried to inherit from `MapLayer`, but I decided in the end that I preferred to render the `svg` container in the component, and inherit from `MapComponent` instead.

Happy to open a PR if you like, but as a starting point I'll leave the TS code here.

```TypeScript
import { LatLngBounds, SVGOverlay as LeafletSVGOverlay } from "leaflet";
import React from "react";

import {
  LayerContainer,
  LeafletContext,
  LeafletProvider,
  MapComponent,
  MapLayerProps,
  withLeaflet
} from "react-leaflet";

type Props = {
  attribution?: string;
  bounds: LatLngBounds;
  opacity?: number;
  zIndex?: number;
} & MapLayerProps;

class SVGOverlay extends MapComponent<Props, LeafletSVGOverlay> {
  leafletElement: LeafletSVGOverlay;
  private container: SVGSVGElement;
  private contextValue: LeafletContext;

  constructor(props: Props) {
    super(props);
    this.bindContainer = this.bindContainer.bind(this);
  }

  get layerContainer(): LayerContainer {
    return this.props.leaflet.layerContainer;
  }

  createLeafletElement(props: Props) {
    return new LeafletSVGOverlay(this.container, props.bounds, this.getOptions(props));
  }

  updateLeafletElement(fromProps: Props, toProps: Props) {
    if (toProps.bounds !== fromProps.bounds) {
      this.leafletElement.setBounds(toProps.bounds);
    }
    if (toProps.opacity !== fromProps.opacity) {
      this.leafletElement.setOpacity(toProps.opacity);
    }
    if (toProps.zIndex !== fromProps.zIndex) {
      this.leafletElement.setZIndex(toProps.zIndex);
    }
  }

  componentDidMount() {
    this.leafletElement = this.createLeafletElement(this.props);
    this.layerContainer.addLayer(this.leafletElement, "");

    this.contextValue = { ...this.props.leaflet, popupContainer: this.leafletElement };

    super.componentDidMount();
  }

  componentDidUpdate(prevProps: Props, prevState: {}) {
    super.componentDidUpdate(prevProps, prevState);

    this.updateLeafletElement(prevProps, this.props);
  }

  componentWillUnmount() {
[...]

PR #630, opened 2019-10-16
Add SVGOverlay component

fenech:
Mentioned in #624, this component exists in Leaflet but has no analogue in this library.

A lot of code is duplicated with `MapLayer` - the reason for this is to give the component access to the SVG element that it creates (`this.container`). An alternative could be to create the DOM element in the constructor and call `super({ ...props, container })`.

Added docs and example.


PaulLeCam:
Thanks!
I think creating the container in the `createLeafletElement()` method is a good way.
Could you please address the issues raised in CI so this can be merged?

fenech:
@PaulLeCam updated to fix `flow` errors and amended the commit. I typed the container as `Element` because it looks like flow doesn't have proper [types for SVG elements](https://github.com/facebook/flow/issues/2332).

PaulLeCam:
Thanks!

PR #633, opened 2019-10-28
Add support for doubleClickZoom option's "center" setting

DerGuteMoritz:
The `doubleClickZoom` option also supports the `"center"` setting as e.g. `scrollWheelZoom` does. This patch adds the missing support for it.


PaulLeCam:
Thanks!

PR #634, opened 2019-11-06
new package for react-leaflet-bing

Charmatzis:
react-leaflet-bing is now obsolete. The new package name is react-leaflet-bing-v2


PaulLeCam:
Thanks!

PR #635, opened 2019-11-06
update react-leaflet-google plugin

Charmatzis:
the react-leaflet-google is obsolete. Now, they can use react-leaflet-google-v2 package.


PaulLeCam:
Thanks!

PR #648, opened 2019-12-10
Update plugins.md

CLu1s:
I added a react mark cluster plugin compatible with v2 of ract-leaflet


CLu1s:
It's already in the docs but pointig to npm page

PR #649, opened 2019-12-12
utils/pick should catch falsy values

CaptainLexington:
When updating dynamic `Path` options, only truthy values are picked out of props to be updated. Path options can have valid falsy values, like `fillOpacity: 0` or `bubblingMouseEvents: false`. It is possible to initialize a dynamic property to a falsy value, but after changing it to a truthy one, it cannot be changed back. This PR changes `utils/pick.js` so that it will pick any value that isn't `=== undefined`.

### DIFF
--- a/src/utils/pick.js
+++ b/src/utils/pick.js
@@ -2,7 +2,7 @@
 
 export default function pick(object: Object, keys: Array<string>): Object {
   return keys.reduce((obj, key) => {
-    if (object[key] !== 'undefined') {
+    if (object[key] !== undefined) {
       obj[key] = object[key]
     }
     return obj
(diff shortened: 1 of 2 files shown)


src/utils/pick.js
```diff
@@ -2,7 +2,7 @@
 
 export default function pick(object: Object, keys: Array<string>): Object {
   return keys.reduce((obj, key) => {
-    if (object[key]) {
+    if (object[key] !== undefined) {
```
PaulLeCam: Could you replace this by `typeof object[key] !== 'undefined'` please?

CaptainLexington: Sure thing, good suggestion! https://github.com/PaulLeCam/react-leaflet/pull/649/commits/88db6a509bbd6b2f91ef873a1fd87338d4185aa0

CaptainLexington:
Thank you for being so responsive! We've already pulled the new release into production.

PR #656, opened 2020-01-31
Add react-leaflet-geodesic to plugins.md

dcdourado:
Recent part-wrap of Leaflet.Geodesic to React.
Although it still has only GeodesicLine, I think it would be usefull to people that might need it.


PaulLeCam:
Thanks!

PR #662, opened 2020-02-11
Add `prefix` property to AttributionControl docs.

tormi:
An example of how to disable AttributionControl prefix: `<AttributionControl prefix={false} />`


PaulLeCam:
Thanks for your PR but this is not a dynamic property so it would be misleading to document it as such.

PR #692, opened 2020-04-03
mapRef create ref looks like a typo?

josh-sea:


example/components/events.js
```diff
@@ -20,7 +20,7 @@ export default class EventsExample extends Component<{}, State> {
     },
   }
 
-  mapRef = createRef<Map>()
```
clementallen, line 23: It is correct. The `<Map>` is defining the Flow type of the ref.  https://stackoverflow.com/questions/50076176/how-to-use-flow-with-react-createref

PaulLeCam:
As commented by @clementallen this is the Flow type.

PR #695, opened 2020-04-06
Fix Popup's autoPan prop not updating

### LINKED ISSUE #694: Popup does not update state of props passed in
## Bug report

- [x] Using the supported version of Leaflet (v1.6.0 minimum) and its corresponding CSS file is loaded.
- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v16.8.0 minimum).
- [x] Using the [latest version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://react-leaflet.js.org/docs/en/intro.html) and that you understand the [limitations](https://react-leaflet.js.org/docs/en/intro.html#limitations).

### Expected behavior
Popup is created with autoPan set to false, change state of autoPan to true. Must now autoPan to popup on rerender of map.

### Actual behavior
Popup is created with autoPan set to false, change state of autoPan to true. Does not autoPan to popup on rerender of map.
Toggling the autopan state has no effect, it behaves like the value you give it on first render.

### Steps to reproduce

[Codepen](https://codepen.io/louis9171/pen/WNvVwEa)

If autoPan is initially true, open popup and pan far away and hit the rerender button, popup will come back into view. The opposite happens if it's initially false. 
However, the toggle button has no effect.

louisnow:
Fixes #694

### DIFF
--- a/src/Popup.js
+++ b/src/Popup.js
@@ -34,6 +34,9 @@ class Popup extends DivOverlay<LeafletElement, Props> {
     if (toProps.position !== fromProps.position) {
       this.leafletElement.setLatLng(toProps.position)
     }
+    if (toProps.autoPan !== fromProps.autoPan) {
+      this.leafletElement.options.autoPan = toProps.autoPan !== false
+    }
   }
 
   componentDidMount() {


PaulLeCam:
Thanks for your PR but dynamic props should rely on the public methods provided by Leaflet, not mutating internals with possible unintended side-effects.

PR #704, opened 2020-04-30
Add viewBox and preserveAspectRatio props to SVGOverlay element

### LINKED ISSUE #702: SVGOverlay does not allow setting the viewBox.
As documented on [Leaflet's SVGOverlay](https://leafletjs.com/reference-1.6.0.html#svgoverlay),

> A viewBox attribute is required on the SVG element to zoom in and out properly.

However, react-leaflet wrapper, [SVGOverlay.js](https://github.com/PaulLeCam/react-leaflet/blob/master/src/SVGOverlay.js#L26) creates the SVG element itself and does not allow passing a `viewBox` option.

Could you please add it? 

- [x] Good reason to bypass the template: direct pointer to culprit function.
- [x] Using the latest version of React-Leaflet.
- [x] Make sure you have fully read the documentation and that you understand the limitations.

### Expected behavior

**API**
I believe it should simply respond to a `viewBox` attribute on it:

```html
<SVGOverlay bounds={bounds} viewBox={box}>....</SVGOverlay>
```
Perhaps with something like this. I do not know the internals of option handling of react-leaflet.
```js
if (props.viewBox) { this.container.setAttribute('viewBox', props.viewBox); }
```

**Dynamic or static**
Though it would not be hard to do, it does not seem to be that useful a property to make dynamic. I don't really have an opinion.

**Footnote**
(by the way, while fixing this, `createElementNS` is only half of the job, the svg element should have `this.container.setAttribute('xmlns', "http://www.w3.org/2000/svg");` applied to it too.

spectras:
Simply the suggestions in #702 as an actual pull request.
- Add `viewBox` property.
- Add `preserveAspectRatio` property (it works hands in hands with `viewBox`).
- Set the `xmlns` attribute. Not *required* but cleaner and matches leaflet.js documentation.

Fixes #702.

### DIFF
--- a/example/components/svg-overlay.js
+++ b/example/components/svg-overlay.js
@@ -18,7 +18,7 @@ export default class SVGOverlayExample extends Component<{}> {
           attribution='&amp;copy <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
           url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
         />
-        <SVGOverlay bounds={rectangle}>
+        <SVGOverlay bounds={rectangle} viewBox="0 0 80 80">
           <rect x="0" y="0" width="100%" height="100%" fill="blue" />
           <circle r="5" cx="10" cy="10" fill="red" />
           <text x="50%" y="50%" fill="white">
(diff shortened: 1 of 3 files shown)


spectras:
Thanks for reviewing, I will look into it tomorrow and apply the changes.

spectras:
Done, please have a look and tell me if this is what you had in mind, and if something else needs to be changed.

I also added the two properties to types, and added `viewBox` to the example. Which does fix it (before this PR, the content of the SVG overlay gets distorted when zooming the map).

PR #718, opened 2020-06-09
Wrong type mentioned for Map and Marker props center  in the doc /website 

AdiBev:
Map and Marker component props center type is ```LatLng | LatLngLiteral | LatLngTuple```

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -127,7 +127,7 @@ const CodeSample = (props = {}) => (
       lang="javascript"
       source={`import L from 'leaflet'
 
-const position = [51.505, -0.09]
+const position = {lat:51.505, lng:-0.09}
 const map = L.map('map').setView(position, 13)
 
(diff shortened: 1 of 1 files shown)


alecananian:
What's the issue with the original example? `LatLngTuple` is defined as a tuple of two numbers:

    export type LatLngTuple = [number, number];

AdiBev:
It's the typescript throwing error for this. In index.d.ts file
`export type LatLngTuple = [number, number];`
 this should be `export type LatLngTuple = number[];`

alecananian:
@AdiBev `number[]` would be an array of numbers of any length, while `[number, number]` is an array of only two values, or a tuple.  This is a valid Typescript type: https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple

PaulLeCam:
As commented above, `[number, number]` is an valid position input for Leaflet.
The TS definitions are not defined in this repo so if there is any issue with them please report in the relevant repository.

PR #723, opened 2020-06-28
Add <hr> lines to mark section limits

chbrandt:
Sections (`##`) are not clearly defined when rendered at the website (https://react-leaflet.js.org/docs/en/components), and I think such text markup/splitting helps when reading API docs. That's all proposing here.


PaulLeCam:
Thanks!

PR #733, opened 2020-07-19
New Example Added (Click to create a new marker)

cbeltrangomez84:


PaulLeCam:
Thanks for your PR but I think the logic presented in it is already covered in other examples so it would be redundant.

PR #735, opened 2020-07-22
Minify umd bundle using terser

anajavi:
### What?
This switches rollup to use terser instead of uglifyjs.

The PR targets V3 branch.

### Why?
Terser allows es2015+ syntax in input source code.

Without this the babel needs to transpile code down to es5, which it is currently doing.


PaulLeCam:
Good idea, thanks!

Issue #698, opened 2020-04-20
React Leaflet v3


PaulLeCam:
An early (alpha) version of React Leaflet v3 is available on npm using the `next` tag: `yarn add react-leaflet@next`.
There is still lots to do in terms of testing and documentation but this first version is working so far to run the [basic examples](https://react-leaflet-v3.now.sh/docs/example-popup-marker).

Among the major changes are:
- Full rewrite based on [React Hooks](https://reactjs.org/docs/hooks-intro.html) (no more extending `Component`)
- Replace Flow by TypeScript for type checking
- Drop IE support (Babel config will only target the "last 2 versions" of browsers)
- Switch to the [Hippocratic License](https://firstdonoharm.dev/)
- The `Map` component is replaced by [`MapContainer`](https://react-leaflet-v3.now.sh/docs/api-map#mapcontainer), behaving differently
- Event handlers must be attached using the [`eventHandlers prop`](https://react-leaflet-v3.now.sh/docs/api-components#evented-behavior) rather than `on*` props, with a similar change for [path options](https://react-leaflet-v3.now.sh/docs/api-components#path-behavior) and [WMS params](https://react-leaflet-v3.now.sh/docs/api-components#wmstilelayer)
- The `react-leaflet` package only exposes a [public API](https://react-leaflet-v3.now.sh/docs/api-map) while a new package, `@react-leaflet/core`, is used to provide the [core APIs](https://react-leaflet-v3.now.sh/docs/core-introduction)
- A new documentation website using [Docusaurus v2](https://v2.docusaurus.io/) with live examples, currently published on [react-leaflet-v3.now.sh](https://react-leaflet-v3.now.sh/)

During the alpha versions cycle there is likely to be breaking changes between releases, these versions are only meant to be used by early adopters and contributors who want to help identify and fix bugs.

If you want to help with this release please try to convert components using v2 APIs to v3 and open issues for bugs or use cases that can't be implemented.
Please keep in mind the breaking changes from v2 are intentional for better stability and compatibility with React or Leaflet, so things that may work in a certain way in v2 do not necessarily work the same way in v3.
Thanks!

dylel:
This is an awesome update to the library, i've done a little bit of testing and things are looking great so far! Obviously this is early days but how will custom components work now? i.e. a custom layers control? I'm attempting to implement this and a little lost on the direction i should take, should this type of component leverage the new useMapEvent hook and manage things from there?

anajavi:
[...]

PR #736, opened 2020-07-23
Drop support for IE in v3

anajavi:
### What?
This drops support for IE.

This results in a following current (but moving) browser support:
```json
{
  "android": "81",
  "chrome": "81",
  "edge": "81",
  "firefox": "77",
  "ios": "13.3",
  "node": "12.18.2",
  "opera": "68",
  "safari": "13",
  "samsung": "11.1"
}
```

### Why?
The issue #698 noted that support for IE will be dropped in v3. However the babel config is using "last 2 versions" which includes both IE 10 and 11.

I think it would be better to explicitly declare supported browsers:
```json
"browsers": [
  "Chrome >= 73",
  "ChromeAndroid >= 73",
  "Firefox >= 68",
  "Edge >= 18",
  "Safari >= 12",
  "Samsung >= 10",
  "iOS >= 12"
]
```

Without explicit browser support almost all users of react-leaflet need to transpile down to their set of browsers.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -12,7 +12,12 @@
             "modules": "commonjs",
             "targets": {
               "node": "current",
-              "browsers": "last 2 versions"
+              "browsers": [
+                "last 2 versions",
+                "not dead",
+                "not ie > 0",
+                "not ie_mob > 0"
+              ]
             },
             "useBuiltIns": "usage"
           }
@@ -35,7 +40,12 @@
             "modules": false,
             "targets": {
               "node": "current",
-              "browsers": "last 2 versions"
+              "browsers": [
+                "last 2 versions",
+                "not dead",
+                "not ie > 0",
+                "not ie_mob > 0"
+              ]
             },
             "useBuiltIns": "usage"
           }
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Thanks!
The idea of dropping support for IE is both to hopefully reduce the amount of code that needs to be transpiled and having to investigate weird issues caused by an outdated browser.
I agree it would be better to have an inclusion list of supported browsers rather than exclusion one as you suggested, is there a way to automatically have the last 2 versions of these browsers rather than having to manually update these?

anajavi:
> I agree it would be better to have an inclusion list of supported browsers rather than exclusion one as you suggested, is there a way to automatically have the last 2 versions of these browsers rather than having to manually update these?

I've modified the list to the following:
```js
"browsers": [
  "last 2 Chrome versions",
  "last 2 Edge versions",
  "last 2 Firefox versions",
  "last 2 iOS versions",
  "last 2 Opera versions",
  "last 2 Safari versions",
  "last 2 Samsung versions"
]
```

I think that the "useBuiltins: 'usage'" should be removed too as the only polyfill it adds is web.dom-collections.iterator, which seems to be unneeded anyway(?). This drops UMD bundle size by some 20kB.

Also: With these browsers removing "@babel/plugin-transform-runtime" actually makes the UMD bundle further 10kB smaller.

PaulLeCam:
Nice, thanks!
Yes if you have additional suggestions to improve the build please don't hesitate to make the changes here, or would you rather I merge this PR now?

anajavi:
> Yes if you have additional suggestions to improve the build please don't hesitate to make the changes here, or would you rather I merge this PR now?

I like smaller more focused PRs, so lets merge this now.

PR #738, opened 2020-07-28
Update intro.md

ericgoe:
- Added a link to Next's dynamic import. Using this method it is possible to use Leaflet-React with Next.


PaulLeCam:
Thanks but this is a generic introduction to the library, I'd rather not get into ecosystem specifics on this page.

PR #739, opened 2020-07-28
Drop @babel/runtime and polyfilling

anajavi:
### What?
This PR removes runtime transform and polyfilling from babel config.

### Why?

1. The runtime and polyfills are not needed as only last 2 versions of browsers are supported
2. It saves bundle size. The UMD bundle is now only 11kB, down from 44kB.

### What else?

The docusaurus build seems to be failing. It also fails for me on branch 'next', so it should not be caused by this PR.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,7 +6,6 @@
         [
           "@babel/preset-env",
           {
-            "corejs": 3,
             "loose": true,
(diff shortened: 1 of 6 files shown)


PaulLeCam:
Nice, thanks!

PR #740, opened 2020-07-28
Upgrade rollup plugins

anajavi:
### What?

This upgrades rollup plugins to more recent versions.

### Why?

Just to keep them up to date as they are now released with namespaced names instead.


PaulLeCam:
Nice, thanks!

PR #766, opened 2020-10-01
Leaflet css cdn added

imanaspaul:
To see the map view properly in an application, we also need leaflet CSS.


PaulLeCam:
This is already documented in the setup page.

imanaspaul:
Okay thanks for letting me know.

PR #767, opened 2020-10-07
Fix: Fixed incorrect react example

dipiash:
I'm founded an incorrect example in documentation for `next` (v3) version of library when I research examples.

Before:
```
<MapConsumer>
    ...
<MapConsumer>  {/* Example without slash */}
```

After:
```
<MapConsumer>
    ...
</MapConsumer> {/* Example with slash */}
```


PaulLeCam:
Thanks!

PR #768, opened 2020-10-08
Update plugins.md

slutske22:
Hey Paul, thanks for react-lealfet, I super love it.  I wrote [react-esri-leaflet](https://github.com/slutske22/react-esri-leaflet) for react-leaflet v3, and I followed your instructions on how to reference a plugin for v3.  I just added support to use react-esri-leaflet with react-leaflet v2, so I wanted to add it to this list.  Thanks again!


PaulLeCam:
Thanks!

PR #772, opened 2020-10-19
fix change viewport when crs changed

NikolayBaranovv:
Hello!

In my app i have many map-providers with different CRS. When i change provider, viewport changed incorrectly. I solved that problem and suggest to you apply this fix.
I add in updateLeafletElement this lines:
```
    if (crs !== fromProps.crs) {
      this.leafletElement.options.crs = crs
      this.leafletElement.setView(center, zoom, {animate: false})
      this.leafletElement._resetView(this.leafletElement.getCenter(), this.leafletElement.getZoom())
    }
```
function _resetView is need for map children elements.

This pictire before change:
before_change
After change:
after_change

Best regards, Nicolay

### DIFF
--- a/src/Map.js
+++ b/src/Map.js
@@ -137,6 +137,7 @@ export default class Map extends MapEvented<LeafletElement, Props> {
       boxZoom,
       center,
       className,
+      crs,
       doubleClickZoom,
       dragging,
       keyboard,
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Thanks but unfortunately applying this kind of logic can have unexpected consequences for other Leaflet plugins for example, or become a breaking change in a future version of Leaflet. That's why as much as possible React Leaflet only uses Leaflet's public methods and does not mutate the `options`.
Instead, you could set a unique `key` when you change the `crs` value on your `Map` element to make sure it gets re-created as needed.

PR #773, opened 2020-10-22
Allow react 17 in peerDependencies

anajavi:
### what?

This adds react 17 to react-leaflet@^2 peerDependencies.

### why?

To allow react version overlap with the react-leaflet@^3.


PaulLeCam:
Thanks!

PR #781, opened 2020-11-03
Use standard SPDX license name

nertzy:
See https://spdx.org/licenses/Hippocratic-2.1.html

The SPDX short identifier is `Hippocratic-2.1`.


PaulLeCam:
Thanks!

PR #785, opened 2020-11-13
Document pathOptions for GeoJSON

bb:
This adds documentation for the GeoJSON showing that it has `pathOptions` prop, same as Rectangle et al.

~It's marked Mutable like for the other Components and changing it is actually delegated to the leaflet object, however it does not work as it needs [`resetStyle()`](https://leafletjs.com/reference-1.7.1.html#geojson-resetstyle) to be called.~

~Setting a different `color` or `weight` via `pathOptions` on a `<Rectangle.../>` did neither work for me, so I wonder if it's a bug or if is intended but the documentation is misleading.~

Edit: Nevermind the update issues. It's updating. I was setting only the sub-property not the whole pathOptions. My problem can be reproduced with the third case within in https://codesandbox.io/s/xenodochial-proskuriakova-pytgj?file=/src/Map.js:8695-8722 and is most likely intentional functionality.


PaulLeCam:
Thanks for the PR!
Yes props are compared by reference so if you want to mutate fields in an object, you need to provide a new object.

PR #788, opened 2020-11-17
Add whenDestroyed handler to MapContainer

bb:
This adds the whenDestroyed handler to MapContainer. It complements whenCreated and allows cleaning up to not have memory leaks when dealing with external state.


PaulLeCam:
Thanks for your PR but I'd rather not add functionality that can be implemented at the app level, in this case with `useEffect()` at the same level as the `<MapContainer>`.

PR #807, opened 2020-12-13
Fix a library name typo in install docs

sevazhidkov:
I was going through the docs and for a few minutes I got stuck, not understanding why my library import wasn't working. :)


PaulLeCam:
Thanks, this has been fixed in another commit.

PR #810, opened 2020-12-14
Make Pane component compatible with StrictMode

### LINKED ISSUE #791: Panes with names do not work in React.StrictMode
## Bug report in v3

**Before** opening an issue, make sure to read the [contributing guide](https://github.com/PaulLeCam/react-leaflet/blob/master/CONTRIBUTING.md) and understand this is a bug tracker, not a support platform.

Please make sure to check the following boxes before submitting an issue.\
**Issues opened without using this template will be closed unless they have a good reason not to follow this template.**

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v17.0.0 minimum).
- [x] Using the supported version of Leaflet (v1.7.1 minimum) and its corresponding CSS file is loaded.
- [x] Using the [latest v3 version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://react-leaflet.js.org/docs/start-introduction) and that you understand the [limitations](https://react-leaflet.js.org/docs/start-introduction#limitations).

### Expected behavior

Panes with names can be added to maps in Strict Mode.

### Actual behavior

Adding panes with names results in error:

```
Uncaught Error: A pane with this name already exists: foo-bar
    at createPane (Pane.js:24)
    at Pane.js:48
    at mountMemo (react-dom.development.js:15846)
    at Object.useMemo (react-dom.development.js:16219)
    at useMemo (react.development.js:1532)
    at Pane (Pane.js:48)
    at renderWithHooks (react-dom.development.js:14985)
    at mountIndeterminateComponent (react-dom.development.js:17811)
    at beginWork (react-dom.development.js:19049)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994)
    at invokeGuardedCallback (react-dom.development.js:4056)
    at beginWork$1 (react-dom.development.js:23964)
    at performUnitOfWork (react-dom.development.js:22776)
    at workLoopSync (react-dom.development.js:22707)
    at renderRootSync (react-dom.development.js:22670)
    at performSyncWorkOnRoot (react-dom.development.js:22293)
    at react-dom.development.js:11327
    at unstable_runWithPriority (scheduler.development.js:646)
    at runWithPriority$1 (react-dom.development.js:11276)
    at flushSyncCallbackQueueImpl (react-dom.development.js:11322)
    at flushSyncCallbackQueue (react-dom.development.js:11309)
    at flushPassiveEffectsImpl (react-dom.development.js:23620)
    at unstable_runWithPriority (scheduler.development.js:646)
    at runWithPriority$1 (react-dom.development.js:11276)
    at flushPassiveEffects (react-dom.development.js:23447)
    at react-dom.development.js:23324
    at workLoop (scheduler.development.js:590)
    at flushWork (scheduler.development.js:545)
    at MessagePort.performWorkUntilDeadline (scheduler.development.js:157)
```

### Steps to reproduce

https://codepen.io/zmbc/pen/vYXBwxa

zmbc:
Fixes #791.

This PR makes use of `useEffect` instead of `useMemo` to handle the side effect of creating a Leaflet map pane in the `<Pane>` component. This means that rendering that component does not fail in `React.StrictMode` and also will make it compatible in the future with Concurrent Mode.

Note that there are still several bugs/instances of weird behavior that this does _not_ fix:

- Panes do not actually respond to updated props--name, style, parent pane are all fixed for the life of the component
- Panes without names do not actually contain their children. You can see this bug in the `.tsx.snap` file.
- Panes do not respond to context changes, nor do they pass updated leaflet context onto their children--they always pass the context at the time they were instantiated. Not sure how big of issues this causes.

All of the above could be fixed by more in-depth changes along the same lines, which I am happy to do if this is well received.


PaulLeCam:
Thanks, can you please add tests for nested panes, children components other than `<TileLayer>` and unmounting/mounting again panes with the same name? These are some of the more complex cases the current behavior should handle.

About the immutable props, this is expected, it's safer to make them immutable for consistency. Regarding the parent context I don't think there are changes that should affect the behavior but in case it could be made mutable.
Also the `name` prop is required, it's not supposed to work without.

zmbc:
> About the immutable props, this is expected, it's safer to make them immutable for consistency.

This makes sense to me, and I can't think of a reason why I'd want mutable props, but this should definitely be documented as it is not the standard/expected behavior of React components. I'd be happy to write something, but I wasn't able to find any API documentation on `<Pane>` except for the example--maybe add a note as a comment in that code if there's nowhere else more appropriate?

> Also the name prop is required, it's not supposed to work without.

Ah, I see. Unfortunately, any project not using TypeScript isn't aware of this, because there is no runtime check. There's a few different strategies I see on the web for reflecting TypeScript prop types as `React.propTypes` to add a runtime check--maybe [this is of interest?](https://github.com/milesj/babel-plugin-typescript-to-proptypes)

zmbc:
I've added tests for nested panes as well as unmounting and remounting.

However, I didn't add tests for other children than `<TileLayer>` because they won't render in the test environment. It seems [you have fixed this problem before](https://github.com/Leaflet/Leaflet/issues/6297#issuecomment-499809735)--do you have a newer approach? There's nothing for me to model my tests on, as there are no existing tests for any other kinds of layers.

PaulLeCam:
> > About the immutable props, this is expected, it's safer to make them immutable for consistency.
> 
> This makes sense to me, and I can't think of a reason why I'd want mutable props, but this should definitely be documented as it is not the standard/expected behavior of React components. I'd be happy to write something, but I wasn't able to find any API documentation on `<Pane>` except for the example--maybe I could add a note as a comment in that code if there's nowhere else more appropriate?

This is [already documented](https://react-leaflet.js.org/docs/api-components#props).

> > Also the name prop is required, it's not supposed to work without.
> 
> Ah, I see. Unfortunately, any project not using TypeScript isn't aware of this, because there is no runtime check. There's a few different strategies I see on the web for reflecting TypeScript prop types as `React.propTypes` to add a runtime check--maybe [this is of interest?](https://github.com/milesj/babel-plugin-typescript-to-proptypes)

Also [already documented](https://react-leaflet.js.org/docs/api-components#pane).
Regarding proptypes, React and the community seem to move away from them so I don't want to add support and extra dependencies.

zmbc:
Wow, can't believe I missed that "Child components" page! Sorry for the noise.

anajavi:
I ran into this problem too.

Can I help to move this PR forward somehow?

zmbc:
@PaulLeCam Not sure if you saw my comment above--I believe this is what's blocking:

> However, I didn't add tests for other children than `<TileLayer>` because they won't render in the test environment. It seems [you have fixed this problem before](https://github.com/Leaflet/Leaflet/issues/6297#issuecomment-499809735)--do you have a newer approach? There's nothing for me to model my tests on, as there are no existing tests for any other kinds of layers.

PaulLeCam:
I don't think mocking Leaflet is that useful to really prevent issues so I'm not doing in anymore in v3.
For panes the [docs example](https://github.com/PaulLeCam/react-leaflet/blob/master/packages/website/docs/example-panes.md) has been useful to check the behavior, mostly related to mounting/unmounting.
If you can figure out a good way to get the exact behavior in unit tests that would be great, please add it to your PR!

zmbc:
@PaulLeCam I'm not sure what you're asking me to do.

Any map layers that use SVG will currently fail because [`jsdom`'s SVG support is still a work in progress](https://github.com/jsdom/jsdom/pull/2926).

The only ways I can think of to add tests of SVG children are:

1. Mocking either Leaflet or, more minimally, [createSVGRect](https://stackoverflow.com/a/54384719). Of course, mocking won't actually render the correct SVG so some issues might not be caught and snapshots will not be accurate.
2. Using some kind of unstable/unreleased version of jsdom.
3. Running these tests in an actual headless browser instead of jsdom.

Would you like me to attempt to do one of these three?

Personally, if it were me, I'd just merge this PR without having all those tests since the jsdom+SVG issue is unrelated to panes and at least this PR adds *some* pane tests, where the existing code has none. But it's your call, obviously.

PaulLeCam:
Headless browser tests would be good yes but I don't think that should be needed for this PR.
All I'm concerned about is that this change could break the behavior or other components. With the logic you're changing in effect every pane will first not render, and only setting its element in the effect callback to then re-render, which is a behavior I wanted to avoid.

Hopefully it will work as expected, I'm up for making a release with this but I'll revert it if it breaks anything else.
Can you resolve the conflict please so I can merge the PR?

zmbc:
... I'm confused. Looking at the merge conflicts, I see that [all useMemo instances were replaced by useState](https://github.com/PaulLeCam/react-leaflet/commit/e2e36643150c04f190944ec638a5f267358a9ff4). (This doesn't fix the StrictMode issue, because [useState initialization callbacks are also not safe places to perform side effects](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects).)

I don't understand this. Using `useMemo` to store newContext makes less sense than using `useState`--it is a cached function result, not a mutable state. `useMemo` was not the right tool for the paneElement because *it performed a side effect*, which newContext does not.

I have reverted these changes in the Pane component, but left them in the MapContainer component, in my merge commit.

PaulLeCam:
Thanks!

PR #814, opened 2020-12-14
Fix typo on installation docs

kevinlg:


kevinlg:
This is already addressed in #807

Issue #453, opened 2018-02-28
Marker not appearing for simple example


toby-griffiths:
Please make sure to check the following boxes before submitting an issue.\
**Issues opened without using this template will be closed unless they have a good
reason not to follow this template.**

* [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
* [x] Using a supported version of React and ReactDOM (v15.x or v16.x).
* [x] Using the supported version of Leaflet (v1.3.x) and its corresponding CSS file is loaded.
* [x] Using the [latest stable version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
* [x] The issue has not already been reported.
* [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
* [x] Make sure you have fully read the [documentation](https://react-leaflet.js.org/docs/en/intro.html) and that you understand the [limitations](https://react-leaflet.js.org/docs/en/intro.html#limitations).

### Expected behavior

Add the simple, example map to an App, and expect to see a map with a map marker that can be clicked to open the popup.

### Actual behavior

Map appear, the marker does not (or more accurately the image of the map is broken.  The element exists on the page).

Clicking on the (invisible) marker opens the popup.

It appears that the map marker src attribute is garbled…

[...]

PR #823, opened 2021-01-03
Improve the out-of-the-box setup instructions

kybernetikos:
Working with create-react-app, the instructions didn't work out of the box for three reasons:
1. The instructions didn't explain how to include the leaflet CSS
2. The container had no height
3. The create react app bundler doesn't work with leaflet icons.  See https://stackoverflow.com/questions/60174040/marker-icon-isnt-showing-in-leaflet

Since this is the 'getting started', I think it's better to get it working nicely, even if it's not exactly best practice (e.g. people should set their container height in CSS elsewhere).

### DIFF
--- a/packages/website/docs/start-setup.md
+++ b/packages/website/docs/start-setup.md
@@ -3,15 +3,28 @@ title: Setup
 ---
 
 1. Follow all the steps from the [installation page](start-installation.mdx)
-1. Add the following code to your app and check it displays correctly:
+1. Ensure that the leaflet css is included in your project.  E.g. for a typical babel app:
 
 ```tsx live
-<MapContainer center={[51.505, -0.09]} zoom={13} scrollWheelZoom={false}>
+import 'leaflet/dist/leaflet.css'
+```
+
+3. Now add the following code to your app and check it displays correctly:
+
+```tsx live
+import markerIconPng from "leaflet/dist/images/marker-icon.png"
+import {Icon} from 'leaflet'
+
+// The default bundler configuration will not work with the way leaflet loads
+// its images, so doing this avoids seeing broken image icons.
+const markerIcon = new new Icon({iconUrl: markerIconPng})
+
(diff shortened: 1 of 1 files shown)


silviuburceadev:
Not sure what Paul thinks but as a new user of React Leaflet, your PR would have saved me a few hours. Setting a height in the example isn't bad imho, it's not a ready-to-use-production-grade app, but it's good enough to get you started and you can customize the example to your needs.

kybernetikos:
> Not sure what Paul thinks but as a new user of React Leaflet, your PR would have saved me a few hours.

As you may have guessed, this PR was the result of my first few hours of starting to use leaflet and react-leaflet :-)

PaulLeCam:
Thanks but this is already described in Leaflet's quick start guide, which is clearly [indicated as a prerequisite to using React Leaflet](https://react-leaflet.js.org/docs/start-installation#leaflet-prerequisites).

kybernetikos:
If you follow those instructions, you'll end up including Leaflet in your app twice won't you?

lobaak:
This is definitely useful to include. A lot of the examples in v3 don't work unless you include the icon manually.

madhums:
I agree, these instructions are not that clear. A short explanation of this should be put in the readme 1) about icons 2) about css or atleast a mention pointing to leaflet docs itself.

madhums:
@PaulLeCam would you be willing to add these improvements? I think they would definitely help

madhums (commented):
If you are using webpack5 along with nextjs, your next.config.js should look like this: you need to use the latest asset modules feature.

```js
module.exports = {
  future: {
    webpack5: true,
  },
  webpack: (config) => {
    // optional
    // config.output.assetModuleFilename = 'images/[hash][ext][query]';
    config.module.rules.push({
      test: /\.png/,
      type: 'asset/resource',
    });

    return config;
  },
};
```

Related issue which needs fixing but is closed and locked: https://github.com/PaulLeCam/react-leaflet/issues/453

PR #837, opened 2021-01-25
Fix a bug where some Marker options were ignored on update.

MasterEric:
Before this change, keys other than `position`, `icon`, `zIndexOffset`, `opacity`, and `draggable` were ignored by the `updateMarker` function if they were modified.

With this change, other keys are now checked for updates, and merged into the marker using the `L.Util.setOptions` function.


PaulLeCam:
Thanks but I don't want to add such "catch-all" behavior, updates are only performed using Leaflet's public APIs for the given element.

PR #858, opened 2021-03-10
optional env prop for WMSTileLayer

stijn26:
Extra optional prop 'env'


PaulLeCam:
This doesn't seem necessary.

PR #864, opened 2021-03-31
export default MapContainer

luismendes070:
https://stackoverflow.com/questions/36426521/what-does-export-default-do-in-jsx

https://stackoverflow.com/questions/65604960/react-leaflet-map-container-not-found


PaulLeCam:
Thanks but this is not useful.

PR #870, opened 2021-04-19
Add GeoJsonLayer incl. ts type guards

brussee:
Hi @PaulLeCam,

Thank you very much for your efforts on this library!
I've created a new component: `GeoJsonLayer`, with an interface as close to the existing react-leaflet GeoJSON component as possible, with added functionality to update the content of the Leaflet layer 'the React way' using other existing react-leaflet components.
This PR also includes TS type guards, used in `GeoJsonLayer`, which I haven't released in a separate repo (yet).

Please let me know what you think, thanks!

### DIFF
--- a/packages/react-leaflet/src/GeoJsonLayer.tsx
+++ b/packages/react-leaflet/src/GeoJsonLayer.tsx
@@ -1,0 +1,221 @@
+import {
+  Feature,
+  GeoJSON,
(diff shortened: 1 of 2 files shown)


PaulLeCam:
That seems like a good idea but it's out of the scope of this library, maybe something that could be created as a separate library?

Issue #891, opened 2021-06-06
Can't compile react-leaflet


pkdhital:
## Bug report

Can't compile blank react app after adding react-leaflet. Throws following error:
`
>./node_modules/@react-leaflet/core/esm/path.js 10:41
>Module parse failed: Unexpected token (10:41)
>File was processed with these loaders:
 >* ./node_modules/react-scripts/node_modules/babel-loader/lib/index.js
>You may need an additional loader to handle the result of these loaders.
>|   useEffect(function updatePathOptions() {
>|     if (props.pathOptions !== optionsRef.current) {
>       const options = props.pathOptions ?? {};
>|       element.instance.setStyle(options);
>|       optionsRef.current = options;

`
### Expected behavior

Should compile and be able to use react leaflet

### Steps to reproduce

>npx create-react-app test
>cd test
>npm install leaflet react-leaflet
add  `import { MapContainer, TileLayer, Marker, Popup } from 'react-leaflet'` on src/App.js
>npm start

## Edit
For me https://github.com/PaulLeCam/react-leaflet/issues/891#issuecomment-860152169 this resolved issue, but comments below indicate it's not working for everybody. Please try this and check if it's working for you or not.

MGraefe:
Having the same issue, related thread in create-react-app:

https://github.com/facebook/create-react-app/issues/9468#issuecomment-694191642

Picked up a lot of activity in the last ~20 days. See last comments for remarks from react guys.

vyemialyanchyk:
{code}
./node_modules/@react-leaflet/core/esm/path.js 10:41
Module parse failed: Unexpected token (10:41)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   useEffect(function updatePathOptions() {
|     if (props.pathOptions !== optionsRef.current) {
>       const options = props.pathOptions ?? {};
|       element.instance.setStyle(options);
|       optionsRef.current = options;
{code}
 the same issue

vyemialyanchyk:
`Failed to compile.

./node_modules/@react-leaflet/core/esm/pane.js 2:27
Module parse failed: Unexpected token (2:27)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
| export function withPane(props, context) {
>   const pane = props.pane ?? context.pane;
|   return pane ? { ...props,
|     pane`

[...]

Issue #882, opened 2021-05-18
Error related to babel-loader in the create-react-app project


Khamidullaev:
# Error related to babel-loader in the create-react-app project

I have a project created using create-react-app. Until recently, everything worked well with react-leaflet. But now, when compiling the project, it gives this error:

```sh
./node_modules/@react-leaflet/core/esm/path.js 10:41
Module parse failed: Unexpected token (10:41)
File was processed with these loaders:
 ./node_modules/react-scripts/node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   useEffect(function updatePathOptions() {
|     if (props.pathOptions !== optionsRef.current) {
>       const options = props.pathOptions ?? {};
|       element.instance.setStyle(options);
|       optionsRef.current = options;
```

This operator (??) is supported in the project, but it throws an error in react-leaflet

sonofjavascript:
Similar problem related to code transpilation in  Chrome 73

```
layer.js:18 Uncaught (in promise) SyntaxError: Unexpected token ?
    at Object../node_modules/@react-leaflet/core/esm/layer.js
```

nnatter:
You might be able to fix this by using the `@babel/plugin-proposal-nullish-coalescing-operator` in your project. See https://github.com/PaulLeCam/react-leaflet/issues/883 for why this is happening.

sonofjavascript:
@nnatter if you are using the last version of the `@babel/preset-env` package there is no need to install the `@babel/plugin-proposal-nullish-coalescing-operator`. According to your post https://github.com/PaulLeCam/react-leaflet/issues/879#issuecomment-842239921, setting fixed version on those packages solves the issue.

Thanks

nnatter:
> According to your post [#879 (comment)](https://github.com/PaulLeCam/react-leaflet/issues/879#issuecomment-842239921), setting fixed version on those packages solves the issue.


This is not a solution. It is a workaround that prevents installing the `3.2.0` version of the package. The same error will pop up as soon as a new version of the package is released.

dennisat:
## Full solution for CRA without eject using the `customize-cra`

1. Add the customize-cra
`yarn add --dev customize-cra`
2. Create the `config-overrides.js` file are root folder with this content
```
const {
  override,
  addExternalBabelPlugins,
} = require("customize-cra");

module.exports = {
  webpack: override(
    ...addExternalBabelPlugins(
[...]

Issue #881, opened 2021-05-18
You May need additional loader to handle the result of these loaders


samuelsodhs:
## Bug report

**Before** opening an issue, make sure to read the [contributing guide](https://github.com/PaulLeCam/react-leaflet/blob/master/CONTRIBUTING.md) and understand this is a bug tracker, not a support platform.

Please make sure to check the following boxes before submitting an issue.\
**Issues opened without using this template will be closed unless they have a good reason not to follow this template.**

- [ ] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [ ] Using a supported version of React and ReactDOM (v17.0.0 minimum).
- [ ] Using the supported version of Leaflet (v1.7.1 minimum) and its corresponding CSS file is loaded.
- [ ] Using the [latest v3 version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [ ] The issue has not already been reported.
- [ ] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [ ] Make sure you have fully read the [documentation](https://react-leaflet.js.org/docs/start-introduction) and that you understand the [limitations](https://react-leaflet.js.org/docs/start-introduction#limitations).

### Expected behavior

Please describe.

### Actual behavior

Please describe.

### Steps to reproduce

Please provide the simplest example possible to reproduce the issue, based on [this CodePen](https://codepen.io/PaulLeCam/pen/xxOgovP).

samuelsodhs:
./node_modules/@react-leaflet/core/esm/path.js 10:41
Module parse failed: Unexpected token (10:41)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   useEffect(function updatePathOptions() {
|     if (props.pathOptions !== optionsRef.current) {
>       const options = props.pathOptions ?? {};
|       element.instance.setStyle(options);
|       optionsRef.current = options;

Sata51:
In my case, this has removed the issue https://github.com/facebook/create-react-app/issues/9468#issuecomment-694191642

anthify:
If you're using Next.js the only way I've been able to get it to work is by using [`next-transpile-modules`](https://github.com/martpie/next-transpile-modules)

```javascript
// next.config.js

const withTM = require("next-transpile-modules")([
  "react-leaflet",
  "@react-leaflet/core",
]);

[...]

PR #896, opened 2021-07-06
fix: #891, #883, #882, #881

miccoh1994:
Fixes Issues: 

- [#891](https://github.com/PaulLeCam/react-leaflet/issues/891)
- [#883](https://github.com/PaulLeCam/react-leaflet/issues/#883) 
- [#882](https://github.com/PaulLeCam/react-leaflet/issues/#882)
- [#883](https://github.com/PaulLeCam/react-leaflet/issues/##881) 

Also fixes an issue when using/developing components from external packages by requiring '@react-leaflet/core' as a peer dependecy:

`No context provided: useLeafletContext() can only be used in a descendant of <MapContainer>`

Since @react-leaflet/core, and consequently `useLeafletContext`, are bundled with react-leaflet, components created with core do not have access to the same context as the MapContainer. This is only achievable by excluding @react-leafler/core from the bundle of react-leaflet.

### DIFF
--- a/packages/react-leaflet/package.json
+++ b/packages/react-leaflet/package.json
@@ -42,13 +42,11 @@
     "types/*",
     "umd/*"
   ],
-  "dependencies": {
-    "@react-leaflet/core": "^1.1.0"
-  },
   "peerDependencies": {
     "leaflet": "^1.7.1",
     "react": "^17.0.1",
-    "react-dom": "^17.0.1"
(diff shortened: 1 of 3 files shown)


miccoh1994:
Using the single fix in [#885 ](https://github.com/PaulLeCam/react-leaflet/pull/885)

PaulLeCam:
Thanks but the core package is a direct dependency, not a peer one.

PR #880, opened 2021-05-17
Use `||` instead of `??` for backward compatbilility

alexander-schranz:
fixes #879 not sure if there are other usages of `??`

Issue #877, opened 2021-05-16
Module parse failed: Unexpected token (10:41) in @react-leaflet/core/esm/path.js


carsten-walther:
## Bug report

**Before** opening an issue, make sure to read the [contributing guide](https://github.com/PaulLeCam/react-leaflet/blob/master/CONTRIBUTING.md) and understand this is a bug tracker, not a support platform.

Please make sure to check the following boxes before submitting an issue.\
**Issues opened without using this template will be closed unless they have a good reason not to follow this template.**

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v17.0.0 minimum).
- [x] Using the supported version of Leaflet (v1.7.1 minimum) and its corresponding CSS file is loaded.
- [x] Using the [latest v3 version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [x] The issue has not already been reported.
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://react-leaflet.js.org/docs/start-introduction) and that you understand the [limitations](https://react-leaflet.js.org/docs/start-introduction#limitations).

### Expected behavior

TypeScript should compile successfully.

### Actual behavior

./node_modules/@react-leaflet/core/esm/path.js 10:41
Module parse failed: Unexpected token (10:41)
File was processed with these loaders:
 * ./node_modules/react-scripts/node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   useEffect(function updatePathOptions() {
|     if (props.pathOptions !== optionsRef.current) {
>       const options = props.pathOptions ?? {};
|       element.instance.setStyle(options);
|       optionsRef.current = options;

### Steps to reproduce

- follow https://react-leaflet.js.org/docs/start-setup and run yarn start

Gimhan-minion:
I do have the exact same issue.  @carsten-walther Did you find out a way to fix the bug?

carsten-walther:
No, unfortunately not.

Mimieye:
I have a similar issue about this. But the error occurs in pane.js file
![1取](https://user-images.githubusercontent.com/7856624/118402823-35e1a380-b69e-11eb-9f07-d2f0bb77e405.PNG)

[...]

PR #927, opened 2021-11-17
More beginner friendly way

ellertsmari:
Thank you for a great maps library!

Many of my students are having problems using react-leaflet because of issues reported in #877,  #880 and #883. It seems to be a common problem: https://stackoverflow.com/questions/67552020/how-to-fix-error-failed-to-compile-node-modules-react-leaflet-core-esm-pat and there is even a video about the solution in YouTube with more than 1500 views: https://youtu.be/tFqj-JKYr4M


PaulLeCam:
Thanks but it's been already addressed in previous PRs comments.

Issue #895, opened 2021-07-03
Popup closes if value of onOpen changes (eg. on re-render)


zdila:
## Bug report

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using a supported version of React and ReactDOM (v17.0.0 minimum).
- [x] Using the supported version of Leaflet (v1.7.1 minimum) and its corresponding CSS file is loaded.
- [x] Using the [latest v3 version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [x] The issue has not already been reported.
- [x] Make sure you have followed the [quick start guide](http://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://react-leaflet.js.org/docs/start-introduction) and that you understand the [limitations](https://react-leaflet.js.org/docs/start-introduction#limitations).

### Expected behavior

This is a regression - it is reproducible in react-leaflet 3.2.0 but not in 3.1.0.

Popup should stay open if its `onOpen` value changes.

### Actual behavior

Popup closes if its `onOpen` value changes.

### Steps to reproduce

https://codepen.io/zdila/pen/vYmNewq

```javascript
const [count, setCount] = React.useState(0);

useEffect(() => {
  setInterval(() => {
    setCount(i => i + 1);
  }, 1000);
}, []);

return (
  <Popup onOpen={() => {}}>
    Count: {count}
  </Popup>
);
```

ulken:
You're probably already aware, but the reason is because you're passing a new, anonymous arrow function inline to `onOpen` on each render. Since props are compared by reference by default, this will be considered a prop change and thus trigger a re-render. A re-render in this context basically means replacing the whole DOM element.

Until — if — fixed, a workaround is to memoize the event handler, using `useCallback`. See https://codepen.io/ulken/pen/JjJWKNM.

Not having looked into the underlying architecture of React Leaflet, it feels suboptimal that changing an event handler will cause the DOM element to be replaced entirely, instead of just updating the event handler.

Bigood:
Came across the same issue for `onClose`, I confirm that using `React.useCallback` solves it:
```
const onClose = useCallback(() => {
   ...
}, [])
...
return (
      <Popup 
         onClose={onClose} 
   ...
```

Posting this for SEO, I came across a few posts before finding the right solution.

PR #928, opened 2021-11-17
Fix Popup onClose and onOpen documented as mutable

tomsotte:
Based on https://github.com/PaulLeCam/react-leaflet/issues/895 and a some tests I made, those props are still not mutable and setting an anonymous function as the value breaks the render of the popup content.


PaulLeCam:
These props are mutable, but if you perform state changes you should expect a re-render and therefore a different Popup to be created.

PR #930, opened 2021-11-25
fix: HTTP urls updated to https; OSM attribution link updated

jpstotz:
I updated a lot of attribution links which still used insecure http. Additionally the attribution links were updated to the current OSM recommendation as shown on this page: https://www.openstreetmap.org/copyright

Updated some other URLs from http to https.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 React components for Leaflet maps.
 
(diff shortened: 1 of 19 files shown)


PaulLeCam:
Thanks!

PR #939, opened 2022-01-05
Fix tooltip position prop

simonasdev:
`TooltipProps` had `position` defined, but it never functioned, since `L.Tooltip` does not accept `position` option. So we need to `setLatLng` manually when opening the popup.


povilasDadelo:
I need this

PaulLeCam:
Thanks!

Issue #943, opened 2022-01-10
"Map container is being reused by another instance" error


vardhan03:
Hi @PaulLeCam, Firstly, thanks for this great library :)
I have been using this library for the past 6 months, it is working fine since yesterday. All of a sudden, from this morning, there is an issue with the Map rendering which throws the following error i.e. "Map container being reused by another instance:
![image](https://user-images.githubusercontent.com/43950124/148762253-6a7bbd71-82d5-455d-83b5-baa00715dce7.png)

I have tried to give a unique "key" and "id" to the Map component which wraps up the container from the library, still the same issue.
I have seen the version got updated yesterday but it should not impact our code as we are still with the **3.2.0** version. As per this https://github.com/PaulLeCam/react-leaflet/issues/941, I thought to update the version and update it, but still, the issue persists.

**Note**: For Tile layers, we are using HERE Maps API and API is retrieving Tile layers and all is working.
Can you please let me know what might be the reason to cause the above error, am I missing something?

OlexG:
Update: downgrading to v3.2.2 fixed the issue

vardhan03:
Yes @OlexG, downgrading fixed the issue for me. I had the same ^3.2.0 which caused the update to the library running yarn install in one of our pipelines, I have fixed it to 3.2.0 for now.

renebigot:
Same error here. It seems that you got a boundaries error before this error (so am I). In my app, the map is not showing and I get an error since the app try to use `fitBounds()`.
Downgrading to 3.2.2 resolved the issue.

PaulLeCam:
Please provide the simplest example possible to reproduce the issue, based on [this CodePen](https://codepen.io/PaulLeCam/pen/xxOgovP).

renebigot:
Here it is : [CodePen](https://codepen.io/renebigot/pen/bGommeZ)

renebigot:
Not a good choice, you're right, even if it worked earlier.
Any idea what I should do to be able to call `fitBounds()` (or other) method on the leaflet map object ?

OlexG:
> > Here it is : [CodePen](https://codepen.io/renebigot/pen/bGommeZ)
> 
> In your example, you re-render the MapContainer by setting it the map instance in state so that's expected to be problematic, you're likely creating a loop.

I had a similar issue, but instead set the map instance into a ref (with the same issue), let me see if I can recreate this

OlexG:
[...]

PR #945, opened 2022-01-12
Fix a bug where map is removed too soon

renebigot:
Fix the 3.2.4 bug referred in #943 

‘useEffect(…, [xxx])’ could be called multiple times, so ‘map?.remove()’ may be called too soon.

Fixed it by cleaning up with ‘useEffect(…, [])’ which is called only once, at component destruction.


PaulLeCam:
Thanks but that would prevent the expected unmounting logic as the map instance is not captured. I pushed similar logic in v3.2.5, please try it out and let me know if that solve your problem.

PR #986, opened 2022-04-26
Added steps for adding the Leaflet CSS file

sero-dev:
Added steps for adding the Leaflet CSS file and overriding the height and width properties.

Followed the given instructions from the installation page, and ran into an issue with the map not rendering properly. Found a [Stack Overflow](https://stackoverflow.com/questions/40365440/react-leaflet-map-not-correctly-displayed) post with the solution. Could not find anything in the introduction, installation page or startup page.

image


PaulLeCam:
The installation page explicitly states Leaflet's getting started guide must be followed first.

PR #1028, opened 2022-09-09
Transpile to es2019 to increase backwards compatibility

kozikkamil:
Transpile to es2019 for more backwards compatibility. With es2020, the library preserves nullish coalescing operators in the output. Currently, it's necessary to run CRA 5 to support that, which is not compatible with React 18. For the time being it would be best if this library could work without CRA 5, out of the box.

### DIFF
--- a/.swcrc
+++ b/.swcrc
@@ -4,6 +4,6 @@
       "syntax": "typescript",
       "tsx": true
     },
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Thanks but no, please see previous issues/PRs about this.

PR #1032, opened 2022-09-22
Allow ImageOverlay bounds to be changed on the fly again

hwonyo:
related pr in the past: https://github.com/PaulLeCam/react-leaflet/pull/328

It was supported in the past, but at some point it was missing.


PaulLeCam:
Thanks for your PR, I had to make some changes to ensure the property type is provided but it should be handled now in v4.1.0, please try it out!

PR #1047, opened 2022-11-26
Support changing `url` in TileLayer

goto-bus-stop:
Hi! I have a use case where I change the URL of a tile layer at runtime to change out query parameters depending on user interactions. With core Leaflet, I use the `setUrl()` method. I think it would make sense to expose this in react-leaflet.


PaulLeCam:
Good idea, thanks!

davwheat:
Big thanks for this PR -- before I was just adding/removing `TileLayer` components, but this is waaay more performant! Thank you! 🚀

PR #1048, opened 2022-11-27
Added support for interactive property changes in Path elements

ymoran00:
Path elements (Polyline etc.) have non-mutable interactive property.
I've vhanged it to be mutable - to allow value change of interactive in React.


PaulLeCam:
Thanks for your PR but I don't want to add functionality that is not directly provided by Leaflet's public APIs.

PR #1063, opened 2023-02-27
Bugfix: bounds update in ImageOverlay

hwonyo:
Fix bug when updating bounds for ImageOverlay component


PaulLeCam:
Good catch, thanks!

PR #1064, opened 2023-02-27
Update `interactive` prop

IanMayo:
The underlying leaflet marker object should be updated if `interactive` changes in the props.

### DIFF
--- a/packages/react-leaflet/src/Marker.tsx
+++ b/packages/react-leaflet/src/Marker.tsx
@@ -40,6 +40,9 @@ export const Marker = createLayerComponent<LeafletMarker, MarkerProps>(
     if (props.opacity != null && props.opacity !== prevProps.opacity) {
       marker.setOpacity(props.opacity)
     }
+    if (props.interactive != null && props.interactive !== prevProps.interactive) {
+      marker.setInteractive(props.interactive)
+    }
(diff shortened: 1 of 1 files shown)


PaulLeCam:
Hi, thanks for your PR but I can't find any mention of the `setInteractive` method in Leaflet's docs, could you please point me to the relevant page?

IanMayo:
Hello @PaulLeCam 

I'm afraid I don't have easy access to the leaflet sources - but was following the code pattern in the rest of this method.  It was an attempt to update this property:
image

PaulLeCam:
Yes that can be set as an option, but I'm wondering about the method call you're using here?
I don't want to use any undocumented or internal Leaflet API in React Leaflet to avoid possible breaking changes.

IanMayo:
Aah,
I see my mistake.

Is there alternate way we would update the state of `interactive`?

PaulLeCam:
Yes, if you set a unique `key` on the element you want to change props for, React will re-create it so the options should be picked-up by React Leaflet when creating the new element.

IanMayo:
Thanks for your support @PaulLeCam 

Behold my current workaround:
image

It looks like my workaround is actually the valid solution, though it looks more like a _react-trick_ than formal support for the property.  But, if you do feel this is the valid solution, please feel free to close the PR.

Thanks again :-)
Ian

Issue #1069, opened 2023-03-21
"Map container is already initialized." error happens when component is in Suspense


jisuo:
## Bug report in v4

**Before** opening an issue, make sure to read the [contributing guide](https://github.com/PaulLeCam/react-leaflet/blob/master/CONTRIBUTING.md) and understand this is a bug tracker, not a support platform.

Please make sure to check the following boxes before submitting an issue.\
**Issues opened without using this template will be closed unless they have a good reason not to follow this template.**

- [x] All peer dependencies are installed: React, ReactDOM and Leaflet.
- [x] Using the latest version of React and ReactDOM v18.
- [x] Using the supported version of Leaflet (v1.8.0 minimum) and its corresponding CSS file is loaded.
- [x] Using the [latest v4 version of React-Leaflet](https://github.com/PaulLeCam/react-leaflet/releases).
- [x] The issue has not already been reported.
- [x] Make sure you have followed the [quick start guide](https://leafletjs.com/examples/quick-start.html) for Leaflet.
- [x] Make sure you have fully read the [documentation](https://react-leaflet.js.org/docs/start-introduction) and that you understand the [limitations](https://react-leaflet.js.org/docs/start-introduction#limitations).

### Expected behavior

The error "Map container is already initialized." should not happen when Suspense happens.

### Actual behavior

The error "Map container is already initialized." happens when Suspense happens.

### Steps to reproduce
Code example to reproduce using react-query to trigger Suspense state. Click the Button to trigger fake api fetch to trigger Suspense state.

https://codesandbox.io/s/modest-noyce-qy21ix?file=/src/Map.tsx](https://codesandbox.io/s/modest-noyce-qy21ix?file=/src/Map.tsx:1017-1126)

jisuo:
As a workaround I added a `key` prop to `<MapContainer key={data.imageUrl}>` so whenever the data.imageUrl is updated and a Suspense is triggered the key prop will destroy the instance and recreate it. Maybe this workaround will help others until there are additional ref-checks added to the MapContainer to prevent the error.

This useCallback in MapContainer is run again on Suspense and since `context` is null it causes it to re-create the map in the same div:

```
const mapRef = useCallback((node)=>{
        if (node !== null && context === null) {
            const map = new LeafletMap(node, options);
            if (center != null && zoom != null) {
                map.setView(center, zoom);
            } else if (bounds != null) {
                map.fitBounds(bounds, boundsOptions);
            }
[...]

PR #1073, opened 2023-04-19
Fixes issue with Map Container is already initialized

mistenkt:
Was running into this error quite frequently when using this library on a `Next` project where our maps were being loaded using `next/dynamic` to prevent SSR. 

Previous suggestions about adding keys to the `MapContainer` component were not working.

Noticed that the code for checking if we have an existin context is not actually updated if/when the context changes, and instead explicitly ignoring the lint warnings for this. `// eslint-disable-next-line react-hooks/exhaustive-deps`.

This fixes the problem.

Related issue: https://github.com/PaulLeCam/react-leaflet/issues/1069

### DIFF
--- a/packages/react-leaflet/src/MapContainer.tsx
+++ b/packages/react-leaflet/src/MapContainer.tsx
@@ -67,7 +67,7 @@ function MapContainerComponent<
       setContext(createLeafletContext(map))
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+  }, [context])
 
   useEffect(() => {
     return () => {


marsy-invests:
Thanks @mistenkt - this PR fixes this error for me.  I've patched it locally until this PR is incorporated into a release.

yuzhakovvv:
Any updates here? 👀

natBizitza:
Hi, any update on this PR?😎

mistenkt:
Until this is merged and a new release is created, you can use my branch by adding the following in your package.json

`"react-leaflet": "https://github.com/mistenkt/react-leaflet/archive/refs/tags/react-leaflet-v4.2.1-gitpkg.tar.gz",`

we've been using it in prod for a few months now without any issues.

natBizitza:
> Until this is merged and a new release is created, you can use my branch by adding the following in your package.json
> 
> `"react-leaflet": "https://github.com/mistenkt/react-leaflet/archive/refs/tags/react-leaflet-v4.2.1-gitpkg.tar.gz",`
> 
> we've been using it in prod for a few months now without any issues.

Thank you! It works very well for me!

tswacast:
This fixed the issue for us as well. If we can do anything to help get this PR merged into the new release, let us know. Much appreciated.

stephanmantler:
This change didn't fix the issue for me, but another small tweak did. Although I am pretty sure it is a crude symptomatic solution and not a proper fix, thus no PR. But I basically just ensured `_leaflet_id` was undefined before calling the constructor:

    // MapContainer.js 
    28   const mapRef = useCallback((node)=>{
    29      if (node !== null && context === null) {
    30           // --- ensure LeafletMap does not get confused by id set in a previous render
    31           node._leaflet_id = null;
    32           const map = new LeafletMap(node, options);

(Next.js 15.0.2, React 18.3.1)

PR #1086, opened 2023-07-14
Support dynamic changing of Marker `interactive` property

someonewithpc:


someonewithpc:
@PaulLeCam Could you take a look at this?

PR #1092, opened 2023-09-04
Allow customizing which class gets used for `LayersControl`

someonewithpc:
My company wants to use images inside the `LayersControl`, so this is intended to allow customizing the underlying `L.Control` class that gets instantiated for the `<LayersControl>` provided by this library. Used as 

```
const MyLayersControl = CustomizableLayersControl(class extends L.Control.Layers {
  // ....
  // You'll likely want to override at least `_initLayout` and `_update`
});
```

See the added test for an example


PaulLeCam:
Thanks but this is something that should be provided as a third-party library

PR #1112, opened 2023-11-29
Update LICENSE.md

DanOrsborne:
Please change your licence type as the Hypocratic licence keeps getting flaged in our IT audit


worc:
Older—and less buggy—versions of this library still at least have an MIT License. And this new license seems like a complete walk away from actually open community standards and permissions. Including an arbitration clause is pretty extreme for a javascript library.

PR #1126, opened 2024-02-09
Can't compare array with operator !==

g45t345rt:
The condition is already true and always force render of markers.
We should compare the array or object with toString() function.


Danielv123 (approved):
Makes sense

PR #1131, opened 2024-04-18
feat(react-leaflet): Improved polyline updates when props changes

gallayl:
As I want to update some other properties in the same Polyline leaflet instance as positions, I think it would be a good idea to improve the `<Polyline />` component.
I've added some commonly used properties to the `updatePolyline(...)` method.


PaulLeCam:
Thanks but that should be provided as a third-party library

